PR #46, opened 2025-12-10
Add explicit reference to workflow subdirectory in plugin manifest

airtwo:
Commands are in commands/workflows/ but that path isn't registered in the plugin manifest. Claude Code doesn't automatically discover commands in subdirectories - it only looks in commands/.


airtwo:
I reinstalled the plugin and looks like it works fine now without this, so I'm closing it.

PR #80, opened 2026-01-15
docs: Add LSP prioritization guidance for project CLAUDE.md

kevinold:
## Summary

- Adds "Customizing Agent Behavior" section to plugin README
- Provides example CLAUDE.md snippets for LSP prioritization
- Documents when to use LSP vs grep for code understanding

## Context

Instead of modifying plugin agents to prioritize LSP, this approach lets projects opt-in by adding guidance to their own CLAUDE.md. This is cleaner because:

1. No plugin changes needed
2. Opt-in for projects that want it
3. Customizable per project's language/tooling

## Test plan

- [ ] Verify README renders correctly on GitHub
- [ ] Test that adding the snippet to a project's CLAUDE.md influences agent behavior

### DIFF
--- a/plugins/compound-engineering/README.md
+++ b/plugins/compound-engineering/README.md
@@ -202,25 +202,33 @@ Prefer LSP operations for semantic code understanding; use grep for text pattern
 **Detailed version:**
 
 ```markdown
-## Code Search Strategy
+## Code Search Strategy (MANDATORY)
 
-When searching or analyzing code, prioritize LSP tools over grep/search:
+**This applies to ALL code search operations, including work delegated to sub-agents.**
+
+When searching or analyzing code - whether directly or via Task tool agents - prioritize LSP tools over rg:
 
 1. **LSP First** - Use LSP operations for semantic code understanding:
+   - `documentSymbol` - List all symbols in a file (ALWAYS START HERE)
    - `findReferences` - Find all usages of a symbol
    - `goToDefinition` - Navigate to where a symbol is defined
-   - `documentSymbol` - List all symbols in a file
    - `hover` - Get type info and documentation
    - `incomingCalls` - Find callers of a function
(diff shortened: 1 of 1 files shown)


kieranklaassen:
Thanks! Should we just reach for LSPs instead of grep either way? If so, maybe we can bake it into the main loop? Curious to hear what you think and how it works for you.

kevinold:
Honestly I’m still testing this. It hasn’t returned the value I’d hoped and I’m not sure if it is the updates or Claude Code shipping flakiness with updates around this. IMHO we should aim to ship both, detect LSPs and use them and fallback to grep. And there are several situations where grep is the answer, so it’s not an either/or, but both/and with preference to LSP if available. I will keep testing this and following up this week. On Jan 20, 2026, at 6:36 PM, Kieran Klaassen ***@***.***> wrote:﻿kieranklaassen left a comment (EveryInc/compound-engineering-plugin#80)
Thanks! Should we just reach for LSPs instead of grep either way? If so, maybe we can bake it into the main loop? Curious to hear what you think and how it works for you.

kieranklaassen:
Hows this working? I still prefer to make the main plugin better than just adding docs. Hows this worikgn for you

kieranklaassen:
Thanks for the thoughtful write-up on LSP prioritization! This overlaps with #124 which adds a formal setup/config system. We're going to hold off on both for now while we think through the right approach to project-level customization. Appreciate the contribution!

kevinold:
Yea I haven't seen it engaging much honestly with LSP, so I defer to your judgement on this. Maybe we close this and pursue another option?

kieranklaassen:
Thanks for this contribution\! We're going to close this in favor of making the LSP prioritization work directly in the agents rather than requiring users to add documentation to their CLAUDE.md. Appreciate the idea though — the opt-in approach is solid thinking.

PR #96, opened 2026-01-21
Fix broken markdown table in dhh-rails-style skill

ashwin47:
## Summary
- Remove quotes from the routing table's Response column to fix markdown table rendering on GitHub

CleanShot 2026-01-21 at 05 38 34@2x

## Before
The table rendered incorrectly because quotes in the first column confused the markdown parser.

## After
Clean table without quotes renders correctly.

CleanShot 2026-01-21 at 05 39 17@2x


kieranklaassen:
Thanks!

PR #99, opened 2026-01-21
docs: add skillkit as multi-agent installation option

rohitg00:
This pull request updates the `README.md` to provide improved installation instructions for the Compound Engineering Plugin. The main enhancement is the addition of a new installation method using SkillKit, which allows users to install and synchronize the plugin across multiple AI agents. The documentation now also explains the advantages of using SkillKit and clarifies the available installation options.

**Installation instructions update:**

* Added a new section describing how to install the plugin using SkillKit, including commands for multi-agent installation and a summary of SkillKit's key benefits.
* Clarified the existing method for installing the plugin directly in Claude Code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,25 @@ A Claude Code plugin marketplace featuring the **Compound Engineering Plugin** â
 
 ## Install
 
+### Option 1: SkillKit (Multi-Agent Support)
+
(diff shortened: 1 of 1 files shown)


kieranklaassen:
Thanks this is cool, but compound engineering is more than just the skills. I think without the commands and the workflows it kind of misses the full point. I'm working on support for more places though!

Thanks for you PR

PR #100, opened 2026-01-21
feat(workflows:plan): Add smart research decision logic

tmchow:
## Summary

- Makes `/workflows:plan` smarter about when to run external web research
- Previously always ran all 3 research agents regardless of task complexity
- Now repo research runs first, then decides on external research based on context

This will make it more automatic to want to use the compound engineering workflows instead of thinking whether or not the task at hand warrants it.

Note: In a follow-up PR I'm going to suggest addition of a `/workflows:brainstorm` step that optionally precedes the plan step but that's a much larger proposal. Will be curious to what @kieranklaassen will think about that one ;) 

## Changes

**Step 0 (Idea Refinement):** Now gathers signals about user familiarity, intent, topic risk, and uncertainty level.

**Step 1 (Repo Research):** Always runs (fast, local). Findings inform the research decision.

**Step 1.5 (Research Decision):** New step that decides on external research:
- High-risk topics (security, payments, external APIs) → always research
- Strong local context (good patterns, CLAUDE.md guidance) → skip external
- Uncertainty or unfamiliar territory → research

**Step 1.5b (External Research):** Conditional - only runs when decision indicates value.

**Step 1.6 (Consolidate Research):** New step to consolidate findings from all applicable research.

### DIFF
--- a/plugins/compound-engineering/commands/workflows/plan.md
+++ b/plugins/compound-engineering/commands/workflows/plan.md
@@ -29,37 +29,65 @@ Before running research, refine the idea through collaborative dialogue using th
 - Focus on understanding: purpose, constraints and success criteria
 - Continue until the idea is clear OR user says "proceed"
 
+**Gather signals for research decision.** During refinement, note:
+
+- **User's familiarity**: Do they know the codebase patterns? Are they pointing to examples?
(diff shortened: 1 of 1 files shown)


kieranklaassen:
you are reading my mind and finishing my sentences TY

PR #150, opened 2026-02-05
fix(compound): prevent subagents from writing intermediary files

tmchow:
## Summary

The `/workflows:compound` command was inconsistently creating intermediary files during the research phase instead of having subagents return text data to the orchestrator.

## Problem

When executing `/workflows:compound`, sometimes subagents would write files like `context-analysis.md`, `solution-draft.md` during research, cluttering the repo with artifacts that should never exist. Other times it worked correctly.

## Root Cause

- The original command listed all 6 subagents as "(Parallel)" including the Documentation Writer
- There was no explicit instruction preventing research subagents from writing files
- The Documentation Writer couldn't actually run in parallel—it needs research results first

## Changes

- Add `<critical_requirement>` block explicitly forbidding subagent file writes
- Restructure into clear two-phase orchestration:
  - **Phase 1**: Parallel research (subagents return text data)
  - **Phase 2**: Sequential assembly (orchestrator writes single file)
  - **Phase 3**: Optional parallel enhancement agents
- Remove Documentation Writer as a subagent (orchestrator handles this in Phase 2)
- Add "Common Mistakes to Avoid" table for quick reference
- Update success output to reflect new structure

## Testing

Run `/workflows:compound` after solving a problem and verify:
- [ ] No intermediary files created during research
- [ ] Only one file written: `docs/solutions/[category]/[filename].md`
- [ ] Phase 1 subagents complete before Phase 2 starts

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -11,8 +11,8 @@
   "plugins": [
     {
       "name": "compound-engineering",
-      "description": "AI-powered development tools that get smarter with every use. Make each unit of engineering work easier than the last. Includes 28 specialized agents, 24 commands, and 15 skills.",
-      "version": "2.28.0",
+      "description": "AI-powered development tools that get smarter with every use. Make each unit of engineering work easier than the last. Includes 29 specialized agents, 25 commands, and 16 skills.",
+      "version": "2.30.0",
       "author": {
         "name": "Kieran Klaassen",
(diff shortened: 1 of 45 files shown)


kieranklaassen (approved):
TY @tmchow\! This needs a quick rebase on main to resolve a small conflict (one line in the Context Analyzer section). After that it's good to merge.

PR #157, opened 2026-02-08
Rewrite workflows:review with context-managed map-reduce architecture

Drewx-Design:
## Summary

- Replaces the v1 review command with a **phased file-based map-reduce** architecture that prevents context overflow
- Sub-agents write full analysis to `.review/` on disk and return only ~100 token summaries to the parent, keeping parent context under ~12k tokens (vs ~30-50k in v1)
- Adds new phases: **PR Intent Analysis** (shared context), **Validation** (catches silent failures), **Judge** (dedup + hallucination removal), and **Deep Analysis** (P1/P2 enrichment)
- Smart agent selection based on PR content instead of running all 13+ agents on every PR
- Cross-platform safe: uses project-relative `.review/` instead of `/tmp/` (fixes Windows path issues)

## Key Architecture Changes

| Phase | What Changed |
|-------|-------------|
| Intent (new) | Single agent produces shared context so specialists don't redundantly re-derive PR intent |
| Map | Agents write JSON to `.review/`, return 1 sentence each (~100 tokens vs 2-4k) |
| Validate (new) | Catches silent agent failures + validates JSON schema before judging |
| Judge (new) | Deduplicates, evidence-checks (removes hallucinated findings), ranks top-N |
| Deep Analysis (new) | Enriches P1/P2 with stakeholder impact, failure scenarios, suggested fixes |
| Reduce | Parent reads only `ENRICHED_FINDINGS.json` (~8k tokens max), spawns todo-creation agents |

## Test plan

- [ ] Run `/workflows:review` on a PR with mixed file types (Ruby + JS) to verify agent selection
- [ ] Verify `.review/` directory is created with expected agent output files
- [ ] Confirm parent context stays under ~14k tokens of agent output
- [ ] Test on Windows to verify cross-platform `.review/` path works
- [ ] Verify hallucinated findings (referencing non-existent files) are removed by judge

### DIFF
--- a/plugins/compound-engineering/commands/workflows/review.md
+++ b/plugins/compound-engineering/commands/workflows/review.md
@@ -1,6 +1,6 @@
 ---
 name: workflows:review
-description: Perform exhaustive code reviews using multi-agent analysis, ultra-thinking, and worktrees
+description: Perform exhaustive code reviews using multi-agent analysis with file-based synthesis to prevent context overflow
 argument-hint: "[PR number, GitHub URL, branch name, or latest]"
 ---
 
@@ -64,7 +64,7 @@ Using .review/ inside the project avoids both issues. All tools (Read, Write, Ba
 REVIEW_DIR=".review"
 rm -rf "$REVIEW_DIR"
 mkdir -p "$REVIEW_DIR"
-echo "$REVIEW_DIR/" >> .gitignore 2>/dev/null  # Ensure it's gitignored
+grep -q '.review/' .gitignore 2>/dev/null || echo "$REVIEW_DIR/" >> .gitignore  # Ensure it's gitignored (idempotent)
 
 # Save PR metadata for agents to reference
 gh pr view <number> --json title,body,files > "$REVIEW_DIR/pr_meta.json"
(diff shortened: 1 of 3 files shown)


gregoryfm:
I ran into this issue today. I asked for a PR review, and it hit the limit. I am on the Max 5 plan.

plugins/compound-engineering/commands/workflows/review.md
```diff
@@ -27,500 +39,526 @@ argument-hint: "[PR number, GitHub URL, branch name, or latest]"
...
-- [ ] If ALREADY on the target branch (PR branch, requested branch name, or the branch already checked out for review) → proceed with analysis on current branch
-- [ ] If DIFFERENT branch than the review target → offer to use worktree: "Use git-worktree skill for isolated Call `skill: git-worktree` with branch name
-- [ ] Fetch PR metadata using `gh pr view --json` for title, body, files, linked issues
-- [ ] Set up language-specific analysis tools
-- [ ] Prepare security scanning environment
-- [ ] Make sure we are on the branch we are reviewing. Use gh pr checkout to switch to the branch or manually checkout the branch.
-
-Ensure that the code is ready for analysis (either in worktree or on current branch). ONLY then proceed to the next step.
-
+- [ ] If ALREADY on the target branch → proceed with analysis on current branch
```
XSAM: I wonder why we keep `- [ ]` instead of a numbered list. What is the benefit of using `- [ ]` in this case?

ultraon:
This one is really good and works. It consumes significantly fewer tokens on the main agent.

Drewx-Design:
Closing — the upstream `workflows:review` command has evolved significantly since this was opened (v2.30 → v2.35) and now handles the context management concerns that motivated this rewrite. The default review command is working well in practice.

This PR's key ideas (phased file-based map-reduce, judge dedup, validation phase, intent analysis) may still be worth proposing as targeted improvements to the current upstream command rather than a wholesale replacement. But the 5-version drift makes this PR unmergeable as-is — it would be a complete file conflict.

Architectural patterns from this work influenced the `deepen-plan` v5.0 refactor (PR #178), so the ideas weren't lost.

kieranklaassen:
Great! I really think the Deepin plan is very interesting. I just have to test it out a little bit since it will add some complexity to the plugin. But I like the idea a lot.

PR #160, opened 2026-02-08
Fix crash when hook entries have no matcher

roberto-mello:
## Summary

- Claude Code allows hook entries without a `matcher` field (e.g., `SessionStart` and `SubagentStop` hooks don't need one)
- The OpenCode converter crashes with `undefined is not an object (evaluating 'matcher.matcher.split')` when converting plugins with matcher-less hooks
- Made `matcher` optional in `ClaudeHookMatcher` type and guarded all accesses in `claude-to-opencode.ts`

## Reproduction

```bash
# With a plugin that has SessionStart or SubagentStop hooks without matchers:
bunx @every-env/compound-plugin install ./plugins/beads-compound --to opencode
# ERROR  undefined is not an object (evaluating 'matcher.matcher.split')
```

## Test plan

- [x] All 34 existing tests pass
- [x] Verified conversion succeeds with a plugin containing matcher-less hooks (SessionStart, SubagentStop)

### DIFF
--- a/src/converters/claude-to-opencode.ts
+++ b/src/converters/claude-to-opencode.ts
@@ -209,9 +209,11 @@ function renderHookStatements(
 ): string[] {
   if (!matcher.hooks || matcher.hooks.length === 0) return []
   const tools = matcher.matcher
-    .split("|")
-    .map((tool) => tool.trim().toLowerCase())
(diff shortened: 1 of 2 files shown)


kieranklaassen (approved):
Amazing! TY

Issue #166, opened 2026-02-09
Context limit reached in review agents


kieranklaassen:
![Image](https://github.com/user-attachments/assets/9ee8f16d-f3d7-4823-8b68-f11850ca2188)

seeing this a lot now with opus 4.6 high effort

joshRpowell:
I just started experimenting with this in my CLAUDE.md 
Will report back.

```md
## Model Usage Policy

**Default Model:** Always use Sonnet for all tasks unless explicitly told otherwise or granted permission to escalate.

**When to ask about Opus:**
Before switching to or recommending Opus, ALWAYS ask the user if the task warrants it. Consider asking when:
- Complex multi-step architectural decisions are needed
- Deep reasoning about system design or tradeoffs
- Debugging particularly gnarly, non-obvious issues
- Tasks involving multiple files with complex interdependencies

**For routine tasks, stick with Sonnet:**
- Feature implementation
- Refactoring
- Writing tests
- Documentation
- Bug fixes with clear causes
- Code reviews

When in doubt, explicitly ask: "This task might benefit from Opus for [specific reason]. Would you like me to switch to Opus, or should I proceed with Sonnet?"

jimcru21:
Got this issue of context limit too on deep-plan

<img width="764" height="535" alt="Image" src="https://github.com/user-attachments/assets/476c5ec3-bc75-4e77-86d8-525b1b263cd9" />

havocy28:
I'm also experiencing this error on a regular basis - their start seems to have coincided with the release of Opus 4.6

kieranklaassen:
Yes it's a bug on their end, I'll see what we can do.

kieranklaassen:
## Related upstream issues in Claude Code

This appears to be a known bug in Claude Code, especially since Opus 4.6. Auto-compression is not triggering before the context limit, causing unrecoverable errors. Tracked upstream:

- **anthropics/claude-code#24497** — [Auto-compression not triggering before context limit — forces conversation restart](https://github.com/anthropics/claude-code/issues/24497) (explicitly mentions Opus 4.6)
- **anthropics/claude-code#25011** — [Context limit reached without manual compaction triggering error](https://github.com/anthropics/claude-code/issues/25011) (labeled `has repro`, `bug`, `area:core`, `memory`)
- **anthropics/claude-code#24498** — [Context limit error on new conversation with few messages](https://github.com/anthropics/claude-code/issues/24498)

[...]

PR #124, opened 2026-01-26
Add configurable review agents via .local.md settings

kieranklaassen:
## Summary

Adds per-project review agent configuration using the `.local.md` plugin settings pattern. Run `/compound-engineering-setup` to generate a settings file, or let workflows auto-detect your project type and pick sensible defaults.

### Settings file

```markdown
---
review_agents: [kieran-rails-reviewer, dhh-rails-reviewer, code-simplicity-reviewer, security-sentinel, performance-oracle]
plan_review_agents: [kieran-rails-reviewer, code-simplicity-reviewer]
---

# Review Context

Add project-specific review instructions here.
Focus on N+1 queries — we've had issues in the brief system.
Skip agent-native checks for internal admin pages.
```

### Setup command

`/compound-engineering-setup` detects your project type (Rails, Python, TypeScript, or General) and creates `.claude/compound-engineering.local.md` with appropriate defaults. Edit the file anytime to add/remove agents or add review context that gets passed to all agents.

### Config-aware workflows

`/workflows:review` and `/workflows:work` read agents from the settings file when present. Without it, they auto-detect project type and use defaults — no config required.

Conditional agents (migrations, frontend, architecture, data integrity) stay hardcoded — they trigger based on file patterns in the PR, not user config.

### Schema drift detection

Wires the existing `schema-drift-detector` agent into `/workflows:review` — runs first on any PR touching `db/migrate/` or `db/schema.rb` to catch unrelated schema changes before other DB reviewers run.

## Test plan

- [ ] Run `/compound-engineering-setup` in a Rails project — verify it detects Rails and creates `.local.md` with Rails defaults
- [ ] Run `/compound-engineering-setup` when `.local.md` already exists — verify it asks before overwriting
- [ ] Run `/workflows:review` with `.local.md` present — verify it uses configured agents
- [ ] Run `/workflows:review` without `.local.md` — verify it falls back to auto-detected defaults
- [ ] Run `/workflows:review` on a PR with `db/migrate/` files — verify schema-drift-detector runs
- [ ] Run `/workflows:work` — verify it reads from `.local.md` when present

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #176, opened 2026-02-12
Add --skip-frameworks flag to /workflows:review

marcindyguda:
@kieranklaassen I am an absolute fan of the compound engineering plugin and mindset in general. I am implementing it right now in our own product which is on a completely different stack, and has a huge codebase. I want to extend our PR review specific with your sub-agents, but don't want to invoke your sub-agents directly, because there might be more in the future, but rather would like to exclude the ones who are not relevant for us. 

## Why

The `/workflows:review` command always runs Rails-specific agents (kieran-rails-reviewer, dhh-rails-reviewer, rails-turbo-expert) even on non-Rails codebases. These agents produce irrelevant findings on non-Rails repos while consuming valuable context window. The other generic agents (security, performance, architecture, etc.) are still super valuable regardless of framework. The `--skip-frameworks` flag lets you skip the framework-specific ones, saving context for the agents that matter.

## Summary

- Adds `--skip-frameworks` flag to the `/workflows:review` command
- When passed, skips the 3 framework-specific agents (kieran-rails-reviewer, dhh-rails-reviewer, rails-turbo-expert) so non-Rails codebases get faster, more relevant reviews
- Default behavior (no flag) is unchanged — all agents run as before

## Test plan

- [ ] Run `/workflows:review 123` — all 13 agents should run (unchanged default)
- [ ] Run `/workflows:review 123 --skip-frameworks` — only 10 generic agents should run, framework-specific agents skipped
- [ ] Verify `--skip-frameworks` is stripped from the review target before PR lookup

### DIFF
--- a/plugins/compound-engineering/commands/workflows/review.md
+++ b/plugins/compound-engineering/commands/workflows/review.md
@@ -1,7 +1,7 @@
 ---
 name: workflows:review
 description: Perform exhaustive code reviews using multi-agent analysis, ultra-thinking, and worktrees
-argument-hint: "[PR number, GitHub URL, branch name, or latest]"
+argument-hint: "[PR number, GitHub URL, branch name, or latest] [--skip-frameworks]"
 ---
 
 # Review Command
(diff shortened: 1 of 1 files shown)


kieranklaassen:
Thanks for this — you're solving a real problem! Running Rails-specific agents on non-Rails codebases wastes context.

Good news: **#124** already solves this with a more flexible approach. Instead of a `--skip-frameworks` flag, it adds a `compound-engineering.local.md` settings file in your project root where you configure exactly which agents to run:

```markdown
---
review_agents: [code-simplicity-reviewer, security-sentinel, performance-oracle, architecture-strategist]
---

# Review Context

Our stack is Next.js + Python. Focus on API boundary security and type safety.
```

Running `/compound-engineering-setup` auto-detects your project type and creates the file with smart defaults. No Rails agents for non-Rails projects. And the markdown body gets passed to every agent as additional review context — so you can include stack-specific instructions.

This also helps with the context limit issues in #166, since you're only running the agents that matter for your stack instead of all 13.

Going to close this in favor of #124 — your use case is exactly what it's designed for. Would love your feedback once it lands!

marcindyguda:
nice, thanks! that's going to be a great context saver :)

PR #197, opened 2026-02-18
feature-video: remove hardcoded R2 URL and require public URL verification

zbeyens:
## Problem
The `feature-video` command hardcodes a specific R2 public domain (`pub-404...r2.dev`) and assumes it works for all users.

In practice, users can upload successfully with `rclone` but still get broken demo links in PRs due to:
- different public R2 domains
- public URL disabled/misconfigured
- stale example domain in docs

## What this changes
- Replaces hardcoded public URL assumptions with an explicit `PUBLIC_BASE_URL` variable.
- Adds a required pre-PR validation step:
  - build `VIDEO_URL` and `PREVIEW_URL`
  - `curl -I` both URLs
  - require HTTP `200` for both before updating the PR body.
- Replaces hardcoded URL in markdown example with `<your-public-r2-domain>` placeholder.
- Adds prerequisite note that a public R2 URL must be known.

## Why this is objective and low-risk
- Docs-only change.
- No behavior changes outside the command guidance.
- Prevents false-success uploads that produce broken demo links.

## Validation
- Verified markdown renders correctly.
- Verified command snippets are syntactically valid shell.

### DIFF
--- a/plugins/compound-engineering/commands/feature-video.md
+++ b/plugins/compound-engineering/commands/feature-video.md
@@ -26,6 +26,7 @@ This command creates professional video walkthroughs of features for PR document
 - Git repository with a PR to document
 - `ffmpeg` installed (for video conversion)
 - `rclone` configured (optional, for cloud upload - see rclone skill)
+- Public R2 base URL known (for example, `https://<public-domain>.r2.dev`)
 </requirements>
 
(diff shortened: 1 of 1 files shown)


kieranklaassen:
TY

PR #232, opened 2026-03-08
fix(mcp): add API key auth support for Context7 server

### LINKED ISSUE #153: Context7 MCP server missing API key auth — all requests hit anonymous rate limit when using PLAN
The plugin's `plugin.json` defines the Context7 MCP server with no authentication:

```json
"mcpServers": {
  "context7": {
    "type": "http",
    "url": "https://mcp.context7.com/mcp"
  }
}
```

This means every Context7 request goes unauthenticated, immediately hitting the anonymous rate limit ("Monthly quota exceeded") regardless of the user's actual account status.

**Workaround:** Manually inject API key header into the plugin cache file — works but gets overwritten on updates.

**Suggested fix:** Add header support, either:
1. Env var expansion: `"CONTEXT7_API_KEY": "${CONTEXT7_API_KEY}"`
2. Switch to `/mcp/oauth` endpoint for browser-based auth

Related: upstash/context7#1713

mvanhorn:
## Summary

- Add `x-api-key` header to Context7 MCP server config in `.mcp.json`, reading from `CONTEXT7_API_KEY` env var with an empty default (`${CONTEXT7_API_KEY:-}`) so it remains optional
- Update README Context7 section to document the authentication setup
- Update Known Issues workaround snippet to include the headers config

Closes #153

## Details

Without authentication, all Context7 MCP requests go unauthenticated and immediately hit the anonymous rate limit ("Monthly quota exceeded"). Claude Code supports `${VAR:-default}` environment variable interpolation in `.mcp.json` headers, so users just need to set `CONTEXT7_API_KEY` in their environment to authenticate.

The empty default (`:-`) means the config still parses correctly for users who haven't set the key yet - they'll just continue to use anonymous access.

## Test plan

- [ ] Verify `.mcp.json` is valid JSON after the change
- [ ] Confirm Claude Code resolves `${CONTEXT7_API_KEY:-}` when the env var is set
- [ ] Confirm Claude Code resolves to empty string when the env var is not set (no parse error)
- [ ] With a valid API key, verify Context7 requests succeed without rate limiting

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/plugins/compound-engineering/.mcp.json
+++ b/plugins/compound-engineering/.mcp.json
@@ -2,7 +2,10 @@
   "mcpServers": {
     "context7": {
       "type": "http",
-      "url": "https://mcp.context7.com/mcp"
+      "url": "https://mcp.context7.com/mcp",
+      "headers": {
+        "x-api-key": "${CONTEXT7_API_KEY:-}"
+      }
     }
   }
 }
--- a/plugins/compound-engineering/README.md
+++ b/plugins/compound-engineering/README.md
@@ -53,12 +53,11 @@ Agents are organized into categories for easier discovery.
 | `design-iterator` | Iteratively refine UI through systematic design iterations |
 | `figma-design-sync` | Synchronize web implementations with Figma designs |
 
-### Workflow (5)
+### Workflow (4)
 
 | Agent | Description |
 |-------|-------------|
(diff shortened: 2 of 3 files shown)


kieranklaassen:
thanks!

PR #250, opened 2026-03-10
feat(hooks): add PreCompact hook to auto-trigger /ce:compound

### LINKED ISSUE #95: Add pre-compact hook to auto-trigger /compound workflow
## Feature Request

Add a hook that triggers before context compaction (or session end) to check if there's anything worth compounding, and if so, automatically run the `/workflows:compound` command.

## Motivation

The `/workflows:compound` command captures valuable knowledge - solved problems, learnings, patterns discovered - but requires manual invocation. By the time you think to run it, the context may have already been compacted or the session ended, losing that valuable context.

A pre-compact hook would ensure knowledge gets captured at the optimal moment - when the full context is still available and fresh.

## Proposed Behavior

1. **Trigger:** Before context compaction or session end
2. **Check:** Analyze the session for compoundable content:
   - Problems that were solved
   - Significant debugging sessions
   - Pattern discoveries or learnings
   - Non-trivial fixes with root cause analysis
3. **Action:** If compoundable content is detected, prompt to run `/workflows:compound` (or auto-run based on user preference)

## Acceptance Criteria

- [ ] Hook triggers before context compaction
- [ ] Intelligently detects if session contains compoundable material
- [ ] Respects user preference (prompt vs auto-run)
- [ ] Doesn't trigger for trivial sessions (simple Q&A, quick fixes)
- [ ] Works with the existing `/workflows:compound` command

## Context

Discussed in the Every Discord. This aligns with the core plugin philosophy: "Each unit of engineering work should make subsequent units of work easier—not harder."

Currently, the `/workflows:compound` command has auto-invoke trigger phrases like "that worked" and "it's fixed", but a pre-compact hook would provide a safety net for capturing knowledge before context is lost.

## Related

- `/workflows:compound` command: `plugins/compound-engineering/commands/workflows/compound.md`
- Existing auto-invoke triggers (lines 166-169 in compound.md)

mvanhorn:
## Summary

- Add a `PreCompact` prompt hook that checks if the session contains compoundable knowledge before context compaction
- Uses `type: "prompt"` so the AI evaluates whether `/ce:compound` is warranted - skips silently if nothing significant was solved
- Placed in `plugins/compound-engineering/hooks/hooks.json` which is auto-loaded by the parser's default path convention

Fixes #95

## How it works

When Claude Code is about to compact context, the hook injects a prompt asking the AI to:
1. Check if the session contains a solved problem, debugging insight, or discovered pattern
2. If yes, run `/ce:compound` to capture it in `docs/solutions/`
3. If nothing significant, skip silently (no disruption to user)

This ensures knowledge gets captured at the optimal moment - when full context is still available, before compaction loses the details.

## Test plan

- [ ] Install plugin and work through a debugging session
- [ ] Trigger context compaction and verify the hook fires
- [ ] Verify it skips silently when no significant problems were solved
- [ ] Verify it triggers `/ce:compound` when a problem was solved

### DIFF
--- a/plugins/compound-engineering/hooks/hooks.json
+++ b/plugins/compound-engineering/hooks/hooks.json
@@ -1,0 +1,15 @@
+{
+  "hooks": {
+    "PreCompact": [
+      {
+        "matcher": "*",
+        "hooks": [
+          {
+            "type": "prompt",
+            "prompt": "Before context compaction, check if this session contains a solved problem, debugging insight, or pattern worth documenting. If so, run /ce:compound to capture it before context is lost. If nothing significant was solved, skip silently."
+          }
+        ]
+      }
+    ]
+  }
+}


tmchow:
@mvanhorn The main idea here is good to auto capture knowledge before it's lost. However, a PreCompact hook integration feels like the wrong triggers point.  Lots of sessions have multiple compaction points so it'll be disruptive to flow, and compaction occurring doesn't necessarily mean knowledge is lost. Compaction mid-session might be the wrong time to actually try to compound.

Lastly, if context is about to be compacted, it means context is nearly full. Running `ce:compound` at that point may itself trigger compaction or fail due to limited remaining context.

mvanhorn:
All three points land. PreCompact fires too late in the context lifecycle to safely run ce:compound, and mid-session compaction is noise rather than a signal that knowledge needs saving. Closing this.

PR #241, opened 2026-03-09
refactor(skills): migrate commands to skills directory structure

mvanhorn:
## Summary
Fixes #116

Migrates all 27 command files from `commands/` to `skills/*/SKILL.md` format, as Claude Code 2.1.3+ merged commands and skills into a unified system. All slash commands (e.g., `/ce:plan`, `/lfg`, `/changelog`) continue to work identically since skills with the same `name:` frontmatter create the same slash commands.

## Changes
- Move `commands/ce/*.md` to `skills/ce-*/SKILL.md` (5 core workflow skills)
- Move `commands/workflows/*.md` to `skills/workflows-*/SKILL.md` (5 deprecated aliases)
- Move `commands/*.md` to `skills/*/SKILL.md` (17 utility skills)
- Remove empty `commands/` directory
- Update `plugin.json` and `marketplace.json` descriptions (29 agents, 47 skills)
- Update `CLAUDE.md` directory structure documentation
- Update `deploy-docs` and `generate_command` skill references

## Testing
- `bun test` - all 313 tests pass
- Slash command names preserved via `name:` frontmatter (e.g., `name: ce:plan` in `skills/ce-plan/SKILL.md`)
- Deprecated `workflows:*` aliases preserved (e.g., `name: workflows:plan` in `skills/workflows-plan/SKILL.md`)

This contribution was developed with AI assistance (Claude Code).

PR #251, opened 2026-03-10
fix(targets): nest colon-separated command names into directories

### LINKED ISSUE #226: Opencode under Windows/NTFS doesn't have the ce:* commands
Just pasting what AI told me:

Image

Under Windows, colon should not be used.

mvanhorn:
## Summary

- Fix Windows/NTFS compatibility by splitting colon-separated command names (e.g. `ce:plan`) into nested directories (e.g. `ce/plan.md`) instead of using colons in filenames
- Applied to opencode, droid, and gemini targets
- Matches the approach already used by the Qwen target (`src/targets/qwen.ts:35-41`)

Fixes #226

## Changes

**`src/targets/opencode.ts`**: Split `commandFile.name` on `:` and create nested directories
**`src/targets/droid.ts`**: Same fix
**`src/targets/gemini.ts`**: Same fix (uses `.toml` extension)

### Before
```
commands/ce:plan.md      # Invalid on Windows/NTFS
commands/ce:work.md
commands/ce:review.md
```

### After
```
commands/ce/plan.md      # Valid on all platforms
commands/ce/work.md
commands/ce/review.md
```

Commands without colons (e.g. `deepen-plan`, `triage`) are unaffected.

## Test plan

- [ ] `bun test` passes (312/312 existing tests pass)
- [ ] Install plugin to opencode target on Windows and verify `ce:*` commands appear
- [ ] Verify non-colon commands still work correctly

### DIFF
--- a/src/targets/droid.ts
+++ b/src/targets/droid.ts
@@ -1,5 +1,5 @@
 import path from "path"
-import { copyDir, ensureDir, writeText } from "../utils/files"
+import { copyDir, ensureDir, resolveCommandPath, writeText } from "../utils/files"
 import type { DroidBundle } from "../types/droid"
 
 export async function writeDroidBundle(outputRoot: string, bundle: DroidBundle): Promise<void> {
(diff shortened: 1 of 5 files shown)


tmchow:
@mvanhorn is this even relevant anymore given the restructuring in #241? Or do you think it's still worthwhile for future proofing?

mvanhorn:
Good question! #251 is still relevant - it fixes the *install targets* (opencode, droid, gemini), not the source layout. After #241, the skill `name:` frontmatter still uses colons (e.g., `ce:plan`), and when targets write these to disk, NTFS rejects colons in filenames. #251 splits those into nested directories at install time, matching what the Qwen target already does. Happy to close if you'd prefer a different approach though.

tmchow:
> Good question! #251 is still relevant - it fixes the *install targets* (opencode, droid, gemini), not the source layout. After #241, the skill `name:` frontmatter still uses colons (e.g., `ce:plan`), and when targets write these to disk, NTFS rejects colons in filenames. #251 splits those into nested directories at install time, matching what the Qwen target already does. Happy to close if you'd prefer a different approach though.

Got it. I think this approach works. Take a look at the failing test?

mvanhorn:
Rebased on main - the test failures were from the pre-existing json-config.ts merge order fix that landed after this branch was created. All 314 tests passing now.

mvanhorn:
Extracted `resolveCommandPath(dir, name, ext)` into `src/utils/files.ts` and refactored all 4 targets (opencode, droid, gemini, qwen) to use it. Each target's inline colon-splitting block is now a single call. 1886c74

PR #259, opened 2026-03-13
fix(kiro): parse .mcp.json wrapper key and support remote MCP servers

### LINKED ISSUE #256: Kiro conversion drops MCP servers and hooks
When converting plugins to Kiro format using `convert` or `install` with `--to kiro`, MCP servers and hooks are not included in the output.

**MCP servers:** When a plugin uses `.mcp.json` (instead of defining `mcpServers` directly in `plugin.json`), the converter doesn't parse the file correctly and drops all MCP servers. Tested with the 4 AWS agent plugins from `awslabs/agent-plugins` — all 7 MCP servers across the plugins were skipped.

Additionally, the converter warns "Kiro only supports stdio" and skips HTTP/remote servers, but Kiro now supports remote MCP servers.

**Hooks:** Plugins that define hooks (e.g., PostToolUse) have them skipped during Kiro conversion. The converter warns that Kiro hooks use a different format but doesn't attempt conversion.

**Reproduction:**
```bash
git clone https://github.com/awslabs/agent-plugins.git

# MCP servers get dropped
bunx @every-env/compound-plugin convert ./agent-plugins/plugins/deploy-on-aws --to kiro

# Hooks get skipped
bunx @every-env/compound-plugin convert ./agent-plugins/plugins/aws-serverless --to kiro
```

Skills convert and activate correctly in Kiro, but without the MCP servers they depend on, they can't function.

isadeks:
Fixes #256 (MCP server portion — hooks conversion tracked separately)

**By submitting this pull request, I confirm that my contribution is made under the terms of the MIT license.**

## Summary
- Fix `.mcp.json` parser to handle the `{ "mcpServers": { ... } }` wrapper key format, not just flat server maps
- Add support for remote/HTTP MCP servers (`url` + `headers`) in addition to stdio (`command` + `args` + `env`)
- Remove outdated warning that Kiro only supports stdio
- Servers with neither `command` nor `url` are warned and skipped

## Changes
- `src/parsers/claude.ts` — unwrap `mcpServers` key when present in `.mcp.json` and referenced MCP paths
- `src/converters/claude-to-kiro.ts` — convert `url`-based servers instead of skipping them; remove outdated "Kiro only supports stdio" warning
- `tests/kiro-converter.test.ts` — add test for HTTP server conversion; update skip test for servers with no command or url

## Testing
- `bun test` — all 315 tests pass (955 assertions) across 35 test files    
- Manual: converted `awslabs/agent-plugins` amazon-location-service and deploy-on-aws plugins to kiro format, MCP servers included in output

### DIFF
--- a/src/parsers/claude.ts
+++ b/src/parsers/claude.ts
@@ -159,10 +159,7 @@ async function loadMcpServers(
   const mcpPath = path.join(root, ".mcp.json")
   if (await pathExists(mcpPath)) {
     const raw = await readJson<Record<string, unknown>>(mcpPath)
-    if (raw.mcpServers && typeof raw.mcpServers === "object") {
-      return raw.mcpServers as Record<string, ClaudeMcpServer>
-    }
-    return raw as Record<string, ClaudeMcpServer>
+    return unwrapMcpServers(raw)
   }
 
   return undefined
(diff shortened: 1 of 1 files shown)


src/parsers/claude.ts
```diff
@@ -158,7 +158,11 @@ async function loadMcpServers(
 
   const mcpPath = path.join(root, ".mcp.json")
   if (await pathExists(mcpPath)) {
-    return readJson<Record<string, ClaudeMcpServer>>(mcpPath)
+    const raw = await readJson<Record<string, unknown>>(mcpPath)
```
tmchow: The mcpServers unwrap logic is duplicated. The same if `(raw.mcpServers && typeof raw.mcpServers === "object")` block appears in both `loadMcpServers` (line ~161) and `loadMcpPaths` (line ~239) in src/parsers/claude.ts. This could be a small helper like `unwrapMcpServers(raw)`.  It's small but worth just addressing.

tmchow (changes requested):

PR #277, opened 2026-03-16
fix: codex workflow conversion for compound-engineering

tmchow:
## Summary

This fixes the Codex install path for the compound-engineering workflows and documents the reasoning behind the change.

Before this branch, our Codex adapter had an inconsistent model:

- Claude commands were converted into Codex prompts plus generated delegate skills.
- Claude skills were copied straight into `.codex/skills`.
- Compound workflow entrypoints like `ce:plan` and `ce:work` only existed as copied skills, but their `SKILL.md` content still referenced Claude-style slash handoffs such as `/ce:plan`.

That meant the installed Codex artifacts were structurally inconsistent. We were exposing workflow entrypoints as copied skills, but the skill bodies were still telling Codex to invoke Claude-style commands that did not exist in Codex as such.

## What Was Wrong

The bug was not just a wording issue in `SKILL.md`. The conversion model itself was wrong for compound-engineering workflows:

1. We copied `ce:*` skills verbatim into Codex without transforming their content.
2. We also carried over deprecated `workflows:*` aliases even though they only exist for backward compatibility in this plugin.
3. We were treating the generic Codex converter as if those namespace rules applied to every plugin.
4. In a follow-up review, we also found that personal Claude-home skill parsing had started preferring frontmatter `name` over the skill directory name, which risked changing sync behavior and causing naming collisions.

## Research

I verified the target behavior against both the public docs and the `openai/codex` / `openai/skills` repos before settling on the implementation.

Key findings:

- Codex does support installing skills under `~/.codex/skills`.
- Codex derives the loaded skill name from the skill directory basename, so a copied skill directory named `ce:plan` is a real Codex skill named `ce:plan`.
- Prompts and skills are separate surfaces. Prompts are explicit entrypoints; skills are the reusable implementation units.
- For this plugin specifically, `workflows:*` is deprecated compatibility surface, while `ce:*` is the canonical workflow namespace.

That led to a clearer model for Codex:

- keep canonical workflow implementations as exact-name skills (`ce:plan`, `ce:work`, etc.)
- generate prompt wrappers for the canonical `ce:*` workflow entrypoints
- rewrite copied skill handoffs to those canonical Codex prompt entrypoints
- omit deprecated `workflows:*` aliases from Codex output entirely

### Codex workflow conversion

- Added Codex content transformation utilities for invocation rewriting.
- The Codex converter now applies the workflow prompt-wrapper model only for the `compound-engineering` plugin.
- Canonical `ce:*` workflow skills get Codex prompt wrappers automatically.
- Deprecated `workflows:*` workflow aliases are omitted from Codex output.
- Copied workflow `SKILL.md` files are transformed during Codex install so intra-workflow handoffs point at the canonical Codex prompt surface.

### Safety / scope fix from review

- The `ce:*` / `workflows:*` behavior is now scoped to the compound-engineering plugin instead of the generic converter.
- Other plugins keep their skills untouched, even if they happen to use those namespaces for unrelated reasons.

### Claude home skill parsing fix from review

- Personal skills loaded from `~/.claude/skills` now keep the directory name as the skill name.
- We still parse frontmatter metadata like `description`, `argument-hint`, and `disable-model-invocation`, but we no longer let frontmatter rename the synced skill implicitly.

### Docs

- Added a solution doc capturing the Codex skill/prompt model and the reasoning behind the design.
- Updated the README so local Codex verification uses the repo codepath instead of `bunx`, which would otherwise test the published package instead of local changes.

## Why This Design

This keeps the Codex target aligned with how the plugin actually wants users to think about workflows:

- `ce:*` is the canonical workflow namespace.
- `workflows:*` is legacy compatibility and should not create duplicate Codex entrypoints.
- Prompts are the explicit user-facing entrypoints.
- Skills remain the underlying implementation units.

It also avoids baking compound-engineering assumptions into the generic converter for all plugins.

## Testing

- Added Codex converter tests for:
  - canonical `ce:*` prompt wrappers
  - omission of deprecated `workflows:*` aliases
  - canonical workflow slash-command rewriting
  - non-compound plugins not receiving compound-specific behavior
- Added a Claude-home regression test proving personal skills keep directory-based names even when frontmatter names differ
- Added Codex writer coverage for transformed copied `SKILL.md` content
- Ran `bun test`

## Commits

- `docs: capture codex skill prompt model`
- `fix: align codex workflow conversion`

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Then run `claude-dev-ce` instead of `claude` to test your changes. Your producti
 **Codex** — point the install command at your local path:
 
 ```bash
-bunx @every-env/compound-plugin install ./plugins/compound-engineering --to codex
+bun run src/index.ts install ./plugins/compound-engineering --to codex
 ```
 
 **Other targets** — same pattern, swap the target:
@@ -97,7 +97,7 @@ bun run src/index.ts install ./plugins/compound-engineering --to opencode
 | Target | Output path | Notes |
 |--------|------------|-------|
 | `opencode` | `~/.config/opencode/` | Commands as `.md` files; `opencode.json` MCP config deep-merged; backups made before overwriting |
-| `codex` | `~/.codex/prompts` + `~/.codex/skills` | Each command becomes a prompt + skill pair; descriptions truncated to 1024 chars |
+| `codex` | `~/.codex/prompts` + `~/.codex/skills` | Claude commands become prompt + skill pairs; canonical `ce:*` workflow skills also get prompt wrappers; deprecated `workflows:*` aliases are omitted |
 | `droid` | `~/.factory/` | Tool names mapped (`Bash`→`Execute`, `Write`→`Create`); namespace prefixes stripped |
 | `pi` | `~/.pi/agent/` | Prompts, skills, extensions, and `mcporter.json` for MCPorter interoperability |
 | `gemini` | `.gemini/` | Skills from agents; commands as `.toml`; namespaced commands become directories (`workflows:plan` → `commands/workflows/plan.toml`) |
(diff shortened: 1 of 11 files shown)


kieranklaassen:
Closing this PR as superseded by merged follow-ups.

The Codex workflow conversion work landed in:
- #284 `chore: remove deprecated workflows:* skill aliases`
- #285 `fix: harden codex copied skill rewriting`

That covers the intended Codex workflow conversion plus the follow-up regression hardening that was needed before merge.

PR #286, opened 2026-03-17
feat: add beta skills framework and setup integration

tmchow:
## Summary

- Add `docs/solutions/skill-design/beta-skills-framework.md` documenting the config-driven beta gate pattern for trialing new skill versions with opt-in users
- Update `/setup` skill to manage `beta: true` flag in `compound-engineering.local.md`, including a new Step 3.5 question and confirmation output

## Context

Core workflow skills (ce:plan, deepen-plan) are being rewritten. This PR ships the foundational pattern and config infrastructure so that beta-eligible skills can be safely rolled out to a subset of users. The actual beta skill content (references/beta.md files) will follow in a separate PR.

## Example

For a skill to then have a beta, what you'd do is add this to the very top of the skill after the frontmatter:

```
<!-- BETA GATE -->
MANDATORY FIRST ACTION. Read `compound-engineering.local.md` in the project root.
If `beta: true` is in the YAML frontmatter, STOP — follow ONLY
[beta.md](./references/beta.md) and ignore everything below.
<!-- END BETA GATE -->
```

then the skill has a `./references/beta.md` definition which is the entirely new skill file, minus frontmatter.

### DIFF
--- a/docs/solutions/skill-design/beta-skills-framework.md
+++ b/docs/solutions/skill-design/beta-skills-framework.md
@@ -90,6 +90,21 @@ skills/ce-plan/
 
 The `references/` directory follows existing plugin conventions for supplementary skill files (see AGENTS.md skill compliance checklist — all reference files must be linked with proper markdown links, not bare backtick references).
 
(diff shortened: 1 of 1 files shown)


kieranklaassen:
thought: Instead of making this a beta flag, can we just add another skill and just use that instead to keep things simple? I do like this, but maybe we can keep it more simple. Curious what you think about just adding another plan version and then we just remove and replace it with the old one. If we see it works, it's an easier way to switch between one and the other.

PR #315, opened 2026-03-19
fix: add cursor-marketplace as release-please component

tmchow:
## Summary

- Adds `.cursor-plugin` as a new release-please package (`cursor-marketplace`), independently versioned like the existing `.claude-plugin` (`marketplace`) component
- Seeds the manifest at `1.0.0` matching the current committed value in `.cursor-plugin/marketplace.json`
- Created `cursor-marketplace-v1.0.0` tag and GitHub release at commit `e469ab0` (where the file was introduced) so release-please has a baseline

## Context

The Cursor marketplace was showing a stale version (2.33) because `.cursor-plugin/marketplace.json` was never wired into release-please. Its `metadata.version` has been stuck at `1.0.0` since the file was created. This fix ensures the version is bumped automatically on future releases.

## Test plan

- [ ] After merge, verify release-please picks up cursor-marketplace as a component in the next release PR

### DIFF
--- a/tests/release-metadata.test.ts
+++ b/tests/release-metadata.test.ts
@@ -130,5 +130,6 @@ describe("release metadata", () => {
 
     expect(changedPaths).toContain(path.join(root, "plugins", "compound-engineering", ".cursor-plugin", "plugin.json"))
     expect(changedPaths).toContain(path.join(root, ".claude-plugin", "marketplace.json"))
+    expect(changedPaths).toContain(path.join(root, ".cursor-plugin", "marketplace.json"))
   })
 })


tmchow:
@codex review again

tmchow:
### Code review

Found 1 issue:

1. Missing test assertion for cursor marketplace drift detection. The test "detects cross-surface version drift even without explicit override versions" was updated with a `.cursor-plugin/marketplace.json` fixture seeded with stale data (`description: "old"`), but the assertion only checks that `.claude-plugin/marketplace.json` is in the changed set. It does not assert that `.cursor-plugin/marketplace.json` is also detected as changed, leaving the new sync code path in `metadata.ts` untested. AGENTS.md says "When adding or changing a target, update fixtures/tests alongside implementation rather than treating docs or examples as sufficient proof."

https://github.com/EveryInc/compound-engineering-plugin/blob/ce719005f6f5a9903f3f7d8b1d5435da7db64377/tests/release-metadata.test.ts#L125-L133

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #223, opened 2026-03-02
feat: add /lfg-ext — external delegate LFG (Codex/Gemini/OpenCode in parallel worktrees)

terry-li-hm:
## Summary

Adds `/lfg-ext`, a new workflow command that runs CE planning and review but replaces `/ce:work` swarm with external AI coding tools (Codex, Gemini CLI, OpenCode) running in isolated git worktrees.

## Motivation

Users with Max20 plans burn tokens on CE swarm agents during execution. For teams that already have Codex, Gemini CLI, or OpenCode installed, delegating execution to those tools is free and often faster — while still getting CE's planning depth and review quality.

## How it works

```
/ce:plan → /deepen-plan → [commit plan] → [worktrees + external delegates in parallel] → /ce:review
```

1. Runs `/ce:plan` + `/deepen-plan` as normal
2. Plan is committed (worktrees see git history only, not the working tree)
3. Tasks decomposed by file independence — one worktree per task via `worktree-manager.sh`
4. External tools launched in parallel (backgrounded)
5. Merges worktrees, runs `/ce:review`

## Key design decisions

- **Uses `worktree-manager.sh`** from the existing `git-worktree` skill — no new dependencies
- **Tool availability check** before starting — graceful fallback to `/slfg` if no external tools installed
- **Decision table** for `/lfg-ext` vs `/slfg` — clear guidance on when to use each
- **OpenCode sandbox note** — documents that OpenCode blocks writes outside its worktree (use Codex for those tasks)

## Validated against

A real Rust feature implementation (`.env` file copying + `.gitignore` management in a worktree manager CLI). Codex implemented the feature from a CE plan. 4/4 tests passed, `cargo clippy` clean on first run.

## Gotcha discovered during testing

Worktrees only see committed git history. If `/ce:plan` writes the plan file but it isn't committed before `worktree-manager.sh create`, delegates can't find the plan. Step 3 (commit plan) is explicit in the command for this reason.

## Files changed

- `commands/lfg-ext.md` — new command
- `.claude-plugin/plugin.json` — version 2.38.1 → 2.39.0, commands 22 → 23
- `CHANGELOG.md` — 2.39.0 entry
- `README.md` — commands count + table entry
- `.claude-plugin/marketplace.json` — version + description sync

PR #328, opened 2026-03-20
feat(ce-work): add Codex delegation mode

mvanhorn:
## Summary

Adds an optional "Codex Delegation" section to `ce:work`, parallel to the existing Swarm Mode section. When the user says "use codex for this work", implementation tasks are delegated to `codex exec` while planning, review, and git operations stay in the current agent.

## Why

Users on Max20 plans burn significant context on code implementation. Codex Pro plans include free compute that can handle the writing work. This keeps Claude focused on planning and review.

## Changes

One file changed: `plugins/compound-engineering/skills/ce-work/SKILL.md` (+53 lines)

The new section follows the same structure as Swarm Mode:
- Decision table (when to use vs standard mode)
- Conversational trigger ("use codex for this work")
- 7-step workflow: availability check, credential extraction, prompt assembly, delegation via stdin pipe, diff review, commit, per-task error fallback

Technical details from a battle-tested implementation ([open-source-contributor](https://github.com/mvanhorn/open-source-contributor) runs this pattern in production):
- `command -v codex` for health check (graceful fallback if not installed)
- `echo "$PROMPT" | codex exec ... -` to pipe long prompts via stdin (avoids shell quoting issues)
- `model_reasoning_effort` config key (verified from `~/.codex/config.toml`)
- GH token extraction needed because sandbox blocks keychain access
- Per-task fallback (one Codex failure doesn't disable it for remaining tasks)

## Relation to PR #223

[PR #223](https://github.com/EveryInc/compound-engineering-plugin/pull/223) (lfg-ext) replaces the entire work pipeline with external delegates in parallel worktrees. This PR adds a mode within the existing `ce:work` flow - same relationship as Swarm Mode has to standard mode. The two are complementary.

## Testing

- `bun run release:validate` passes (29 agents, 45 skills, 1 MCP server - no count changes)
- Section is gated behind conversational trigger - standard ce:work is untouched
- Verified `codex exec` invocation pattern works in production via [open-source-contributor commit 588cf99](https://github.com/mvanhorn/open-source-contributor/commit/588cf99)

### DIFF
--- a/plugins/compound-engineering/skills/ce-plan/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-plan/SKILL.md
@@ -151,10 +151,6 @@ Think like a product manager - what would make this issue clear and actionable? 
 - [ ] Identify who will be affected by this issue (end users, developers, operations)
 - [ ] Consider implementation complexity and required expertise
 
-**Execution Posture:**
-
-- [ ] If the user asks for external delegation, says "use codex", "delegate mode", or mentions token conservation, add `Execution note: Execution target: external-delegate` to implementation-heavy tasks in the plan
-
 **Content Planning:**
 
 - [ ] Choose appropriate detail level based on issue complexity and audience


tmchow:
@mvanhorn love this. Is it possible to refactor into a work-beta skill for time being while we test it?

mvanhorn:
@tmchow funny enough I had it in beta, but to simplify my PR I took it out.  let me get it in there

tmchow:
Another interesting thing I wonder is how we might add a config for this so it would use Claude vs codex by default for this phase. But it would have to be careful which coding env you're in and respect the flag. Just thinking out loud.

mvanhorn:
> Another interesting thing I wonder is how we might add a config for this so it would use Claude vs codex by default for this phase. But it would have to be careful which coding env you're in and respect the flag. Just thinking out loud.

interesting! I've only used CE in CC.

tmchow:
> interesting! I've only used CE in CC.

Even just in CC, it would make sense possibly for a setting you set with the `/setup` then you by default use Codex if you want. Otherwise it's on demand. Either way, it's probably fine as is, then we can add that later. Let me look at the skill work you've done.

mvanhorn:
Sounds good - I'll keep it on-demand for now and we can add a `/setup` config option in a follow-up. Let me know what you think after looking at the skill.

tmchow:
@mvanhorn some thoughts after spending time reviewing and thinkign about this.

## Environment Detection

The skill runs across harnesses (Claude Code, Codex, Gemini CLI, etc.). If someone runs `ce:work` from *inside Codex*, this will attempt to shell out to `codex exec` from within Codex's sandbox — likely failing silently or recursing. We need a guard that detects "am I already running inside Codex?" and falls back to standard mode with a clear message. Check for something like a `CODEX_SANDBOX` env var or equivalent before attempting delegation.

## Strategy Integration

The Codex delegation section is bolted on after Swarm Mode as a standalone workflow, but `ce:work` already has a strategy selection step (Phase 1 Step 4) with three modes: inline, serial subagents, parallel subagents. The delegation section doesn't explain how it interacts with these.

Key questions the skill needs to answer:

- If using subagent dispatch, does each subagent shell out to Codex for tagged tasks? (That's Claude Code → subagent → codex exec → GPT — three layers deep)
- If using inline, does the Codex workflow replace the Phase 2 task loop for tagged tasks?
- The 7-step Codex workflow has its own review + commit steps that duplicate Phase 2's incremental commit logic — two parallel "how to execute a task" code paths will drift

**Suggestion**: wire delegation into the existing strategy selection as a task-level modifier rather than a separate section. The strategy (inline/serial/parallel) still applies, but the implementation step within each task delegates to Codex instead of executing directly.

## Plan Abstraction

The `ce:plan` and `ce:plan-beta` changes bake `Execution target: codex` directly into the planning abstraction. Plans should be tool-agnostic — today it's Codex, tomorrow it could be a different delegate. wonde if something like `Execution target: external-delegate` and resolve the specific tool at execution time based on config? This also sets up the `/setup` config option later we discussed earlier

## Hardcoded Model

you use `-m "gpt-5.3-codex"` but that will go stale. This should either come from config (`compound-engineering.local.md`) or but i think omitted entirely to use Codex's default model?

## Mixed Attribution

Phase 4 requires attribution like `Generated with [MODEL] via [HARNESS]`. When half the tasks are written by Codex and half by Claude, what should the attribution say? Needs guidance for mixed-model work... or should we just omit it?

## Error Threshold

"One failure does not disable Codex for remaining tasks" is fine for transient errors, but not for systemic issues (API down, auth expired, rate-limited). Consider a consecutive-failure threshold (e.g., 3 failures in a row → disable delegation for remaining tasks)?

## Beta Skill

Per our earlier discussion, refactor this into a `ce:work-beta` skill with `disable-model-invocation: true` so we can test it safely before promoting.

## Already flagged by automated review (still open)

- `$(cat $CODEX_PROMPT_FILE)` expands into argv, not stdin. `ARG_MAX` risk on larger prompts. Pipe via stdin as described in the step text
- Shell chaining/redirection in examples violates AGENTS.md rules (no `&&`, pipes, `>/dev/null 2>&1` in skill shell examples)

mvanhorn:
Took a stab at this - I like your thoughts. Heads up the description for Compound Engineering on GH says a Claude Code skill.

Addressed all points in 289d146:

**Environment Detection** - Added sandbox guard that checks `CODEX_SANDBOX`, `CODEX_SESSION_ID`, and `.git/` write access before attempting delegation. Falls back to standard mode with a message.

**Strategy Integration** - Rewrote the section to frame delegation as a task-level modifier on the existing strategy (inline/serial/parallel), not a parallel workflow. The strategy still applies - delegation just changes how individual tagged tasks execute.

**Plan Abstraction** - Replaced `Execution target: codex` with `Execution target: external-delegate` across ce:plan, ce:plan-beta, and ce:work. The specific tool resolves at execution time.

**Hardcoded Model** - Removed `-m "gpt-5.3-codex"` flag. Delegates use their own default model.

**Mixed Attribution** - Added guidance for single-delegate, single-standard, and mixed-model scenarios in Phase 4.

**Error Threshold** - Added 3-consecutive-failure circuit breaker that disables delegation for remaining tasks.

**Beta Skill** - Created `ce:work-beta` with `disable-model-invocation: true`.

**Shell compliance + stdin** - Removed all shell chaining/redirection from examples. Prompt goes via stdin, not argv. Credentials via env vars, not prompt text.

tmchow:
@codex provide a final review, focused on the latest changes.  also check if the repo and various READMEs need to be updated.

mvanhorn:
Fixed in 09af36d - ce-work-beta is now a full copy of ce-work with the delegation section integrated, per the beta-skills-framework pattern. Removed the External Delegate Mode section from stable ce-work so it stays untouched.

tmchow:
Nice work @mvanhorn

mvanhorn:
Thanks for the merge!

PR #341, opened 2026-03-22
feat: add OpenPackage support

TiagoJacinto:
Enables installation via opkg install compound-engineering-plugin

Adds openpackage.yml manifest for opkg compatibility.


tmchow:
@TiagoJacinto can you look at the feedback from codex please?

PR #350, opened 2026-03-23
fix: YAML parse error and invalid skill names for pi target

pepegar:
## Bug 1: YAML parse error in frontend-design SKILL.md

The multi-line `description` field in `plugins/compound-engineering/skills/frontend-design/SKILL.md` contains unquoted colons (e.g. `any frontend work: landing pages`) which `js-yaml` interprets as a YAML mapping entry rather than a continuation of the scalar value.

This causes a `bad indentation of a mapping entry` error when running:

```
bunx @every-env/compound-plugin install compound-engineering --to pi
```

```
ERROR  bad indentation of a mapping entry (3:20)

 1 | name: frontend-design
 2 | description: Build web interfaces with genuine desi ...
 3 |   any frontend work: landing pages, web apps, dashb ...
------------------------^
```

**Fix:** Quote the description string so colons are treated as literal characters.

## Bug 2: Invalid skill names for pi target

Pi requires skill names to contain only lowercase `a-z`, `0-9`, and hyphens. The `claude-to-pi` converter was passing skill names through without normalization, producing invalid directory/skill names like `ce:brainstorm` and `generate_command`.

This causes pi to report skill conflicts on startup:

```
[Skill conflicts]
  auto (user) ~/.pi/agent/skills/ce:brainstorm/SKILL.md
    name contains invalid characters (must be lowercase a-z, 0-9, hyphens only)
  ...
```

**Fix:**
- Apply `normalizeName()` to `skillDirs` in `convertClaudeToPi` (was only applied to agents/commands)
- Strip underscores in `normalizeName` (pi doesn't allow them)
- Normalize the `name` field in copied SKILL.md frontmatter to match the pi-safe directory name

### DIFF
--- a/src/converters/claude-to-pi.ts
+++ b/src/converters/claude-to-pi.ts
@@ -36,7 +36,7 @@ export function convertClaudeToPi(
   return {
     prompts,
     skillDirs: plugin.skills.map((skill) => ({
-      name: skill.name,
+      name: normalizeName(skill.name),
       sourceDir: skill.sourceDir,
     })),
     generatedSkills,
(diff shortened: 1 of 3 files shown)


tmchow:
Bug #2 is being fixed as a part of a larger PR for Pi compatibility in #288.

For Bug #1, i just opened #350 so we can fast track that for you

PR #363, opened 2026-03-24
fix(ce-work-beta): add explicit sandbox and approval flags to Codex delegation

mvanhorn:
## Summary

The ce-work-beta delegate step tells the agent to "run the delegate CLI" without specifying sandbox mode or approval policy flags for `codex exec`. This causes delegation to silently fail for users who don't have global yolo config in `~/.codex/config.toml`.

## Why this matters

Codex CLI defaults to a restrictive sandbox that blocks network access and an approval policy that prompts the user. In `exec` (non-interactive) mode, there's no user to prompt, so Codex gives up on any operation that needs approval - like `npm install` or `pip install`.

Users who have `approval_policy = "never"` and `sandbox_mode = "danger-full-access"` in their global `~/.codex/config.toml` never see this because their config overrides the defaults. But anyone with default Codex settings sees delegation fail on the first network operation, triggering the fallback to standard mode every time.

I hit this exact pattern while fixing Codex delegation in another project. A contributor reported that delegation always failed and fell back. Root cause was identical: missing `-s` and `-a` flags on the `codex exec` invocation.

## Changes

Added `-s workspace-write -a never` to step 4 ("Delegate") in `skills/ce-work-beta/SKILL.md`:

- `-s workspace-write` - grants workspace disk writes + network access
- `-a never` - disables approval prompts (required for non-interactive exec mode)

`workspace-write` is the least-privilege option that still allows what delegation needs. It's not `danger-full-access` - Codex can only write within the workspace directory.


tmchow:
superceded by #364. @mvanhorn agrees.

PR #364, opened 2026-03-24
feat(ce-work-beta): add user-selectable sandbox options to Codex delegation

huntharo:
## Summary

Add a security posture selection step to the Codex External Delegation Workflow so users can choose between default, workspace-write, and full-access (`--yolo`) modes before delegation runs.

## Why this matters

Taking over PR #363 after discussing with @mvanhorn .

`codex exec` with default settings fails on the first operation needing approval (e.g., `npm install`) because exec mode is non-interactive. That PR proposed hardcoding `-s workspace-write -a never`, but `workspace-write` still restricts system-level (and network) access outside the workspace directory. For non-trivial one-shot delegations, `--yolo` (`--dangerously-bypass-approvals-and-sandbox`) is the only flag set likely to produce a complete solution -- but it carries real risks (arbitrary file deletion, secret leakage, unapproved command execution).

Rather than hardcoding one posture, this PR lets the user choose with full visibility into the tradeoffs.

## Changes

Added step 2 ("Select security posture") to the External Delegation Workflow in `plugins/compound-engineering/skills/ce-work-beta/SKILL.md`:

- **Default (no flags)** -- Codex defaults. Will almost certainly fail in exec mode.
- **Workspace write** (`--full-auto`) -- `-s workspace-write -a on-request`. Moderate risk, may still fail for tasks needing system access.
- **Full access** (`--yolo`) -- `--dangerously-bypass-approvals-and-sandbox`. Explicit risk warnings about data deletion, secret leakage, and unapproved commands. Only option likely to work for non-trivial exec-mode delegation.

The choice is asked once per session via the platform's blocking question tool (with numbered-list fallback for platforms without one). Renumbered subsequent steps (old 2-7 -> new 3-8).

Supersedes #363.

## Testing

Tested via `claude-code-ce` script that loads updated plugin into claude code CLI, then giving the problem below and checking before/after.

There is an additional oddity uncovered: it tries to redirect the prompt file into `codex` and this fails and it ends up reading help messages several times then deciding to just emit the full prompt in the command line string - other times it randomly decides to cat the plan and pipe it to `codex`, which appears to work.

### Test Prompt

```
/ce:work-beta Run `npm view dive` via Codex.
```

### After - Prompts for Mode

image

### After - Yolo Selected - Runs with `dangerously-bypass-approvals-and-sandbox`

image

### After - Yolo Selected - Gets Result

image

### After - Workspace Write Selected - Runs with `full-auto`

image

### After - Workspace Write Selected - Fails on Network

image

### DIFF
--- a/plugins/compound-engineering/skills/ce-work-beta/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-work-beta/SKILL.md
@@ -470,7 +470,7 @@ When external delegation is active, follow this workflow for each tagged task. D
 
 6. **Review diff** — After the delegate finishes, verify the diff is non-empty and in-scope. Run the project's test/lint commands. If the diff is empty or out-of-scope, fall back to standard mode for that task.
 
-7. **Commit** — The current agent handles all git operations. The delegate's sandbox blocks `.git/index.lock` writes, so the delegate cannot commit. Stage changes and commit with a conventional message.
+7. **Commit** — The current agent handles all git operations. In default and workspace-write modes the delegate's sandbox blocks `.git/index.lock` writes, so the delegate cannot commit. In `--yolo` mode the sandbox is disabled and the delegate *could* commit, but the prompt in step 3 already instructs it not to. Stage changes and commit with a conventional message.
 
 8. **Error handling** — On any delegate failure (rate limit, error, empty diff), fall back to standard mode for that task. Track consecutive failures - after 3 consecutive failures, disable delegation for remaining tasks and print "Delegate disabled after 3 consecutive failures - completing remaining tasks in standard mode."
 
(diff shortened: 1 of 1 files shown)


huntharo:
I ran this last night via the wrapper shell script and built a full stack app in about 45 minutes from Claude using Codex to do the build.

mvanhorn:
@huntharo That's great to hear - full stack app in 45 minutes is a solid validation of the Codex delegation flow. Any rough edges or things that could be smoother?

tmchow:
@huntharo can you rebase off main as i've merged a lot in.  I'd like to figure out a path to get this into the main ce:work skill instead of maintaining it as separate provided we can get enough testing it. Given how big of a boost this is, it's helpful to converge.

One thing to consider is whether we should store a user preference for always running codex mode when the skill is invoked from claude code. remember that we are cross agent env compatible by design so running codex mode from within codex won't make sense.  I'm planning on moving preferences sometime soon out of the `compound-engineering.local.md` and instead moving it to something like `~/.compound-engineering/config.json` (or `config.yaml`). but for now, propose a preference setting that the skill optionally reads from?  it can fallback to conservative if setting is missing and only use codex mode if user invoked skill mentioning it.

huntharo:
> @huntharo can you rebase off main as i've merged a lot in. I'd like to figure out a path to get this into the main ce:work skill instead of maintaining it as separate provided we can get enough testing it. Given how big of a boost this is, it's helpful to converge.
> 
> One thing to consider is whether we should store a user preference for always running codex mode when the skill is invoked from claude code. remember that we are cross agent env compatible by design so running codex mode from within codex won't make sense. I'm planning on moving preferences sometime soon out of the `compound-engineering.local.md` and instead moving it to something like `~/.compound-engineering/config.json` (or `config.yaml`). but for now, propose a preference setting that the skill optionally reads from? it can fallback to conservative if setting is missing and only use codex mode if user invoked skill mentioning it.

Should be all done @tmchow - I have my Codex doing a code review too just in case it finds anything.

tmchow:
@huntharo address the relevant codex feedback

PR #405, opened 2026-03-27
feat(ce-review): add base: and plan: arguments, extract scope detection

tmchow:
## Summary

Adds caller-supplied context arguments to `ce:review` to improve effectiveness of the code review. This is going to be helpful especially later when we want to improve the `/lfg` skill to be a true autopilot.

Two new optional tokens in ce:review's argument parser:

- **`base:<ref>`** — skips the ~150-line scope detection shell when the caller already knows the diff base (SHA, branch name, `origin/main`). Automated callers like lfg/slfg/ce:work benefit most; the full detection remains as fallback for interactive use.
- **`plan:<path>`** — triggers a new Stage 2b that loads the plan's requirements trace and checks each requirement against the diff in Stage 6. Routing depends on how the plan was found: explicit plans (caller-provided or PR body) produce P1/actionable findings; auto-discovered plans produce P3/advisory findings that stay in the report only.

The duplicated branch-mode and standalone-mode base-detection shell (~120 lines repeated twice) is extracted into `references/resolve-base.sh`. Both paths now call the same script.

Workflow callers (lfg, slfg, ce:work, ce-work-beta) updated to record the plan file path from ce:plan and pass it as `plan:<path>` to ce:review.

A `description:` argument was considered and rejected during development — Stage 2's git-log inference is cheap and already has the diff in front of it, so having callers summarize "what was implemented" just moves work upstream without adding signal. The interactive question ("What is the primary goal of these changes?") handles the ambiguous-intent case better than a formatted argument.

---

[Compound Engineering v2.54.1](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/plugins/compound-engineering/skills/ce-review/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-review/SKILL.md
@@ -305,7 +305,7 @@ Pass this to every reviewer in their spawn prompt. Intent shapes *how hard each 
 Locate the plan document so Stage 6 can verify requirements completeness. Check these sources in priority order — stop at the first hit:
 
 1. **`plan:` argument.** If the caller passed a plan path, use it directly. Read the file to confirm it exists.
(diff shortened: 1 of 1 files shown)


tmchow:
@codex review

PR #414, opened 2026-03-27
feat(codex): add native plugin marketplace metadata

mattjefferson:
feat(codex): add native plugin marketplace metadata

## Summary
- add a repo-local Codex marketplace under `.agents/plugins/marketplace.json`
- add native `.codex-plugin` manifests for `compound-engineering` and `coding-tutor`
- document Codex installation in both plugin READMEs
- add coverage for Codex marketplace and manifest alignment

## Testing
- bun test tests/codex-plugin-marketplace.test.ts tests/codex-writer.test.ts tests/codex-converter.test.ts tests/sync-codex.test.ts tests/codex-agents.test.ts
- bun run release:validate

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -1,0 +1,32 @@
+{
+  "name": "compound-engineering",
+  "interface": {
+    "displayName": "Compound Engineering"
+  },
+  "plugins": [
+    {
+      "name": "compound-engineering",
+      "source": {
+        "source": "local",
+        "path": "./plugins/compound-engineering"
+      },
+      "policy": {
+        "installation": "AVAILABLE",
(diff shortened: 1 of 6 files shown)


tmchow:
@mattjefferson thanks for the contribution.. thinking through this because I want to make this work directly for codex officially now that it supports plugins but my understanding is we need to publish the plugin in a different way since codex doesn't support agents in the same way.

mattjefferson:
@tmchow I'm going to close the PR for now and in my testing the plugin system isn't ready yet for this use case.

From OpenAI documentation "Self-serve plugin publishing and management are coming soon."   So it would be better to revisit this next week or so.  And agree will need to think about the Agents in Codex.

PR #441, opened 2026-03-29
feat(resolve-pr-feedback): add gated feedback clustering to detect systemic issues

tmchow:
## Summary

The resolve-pr-feedback skill processes PR comments individually, which leads to a whack-a-mole pattern: fix 3 error handling issues, review bot finds 3 more, repeat. When concentrated feedback in the same area points to a systemic issue, the skill should investigate broadly rather than patching one comment at a time.

This adds a **gated cluster analysis step** that detects when thematically similar feedback in the same code area warrants broader investigation, and dispatches a single agent per cluster to read the wider context before fixing.

## What changed

**SKILL.md** — New Step 3 (Cluster Analysis) inserted between Triage and Plan, with all subsequent steps renumbered (3→4 through 8→9):

- **Gate check**: Two signals — volume (4+ items) or verify-loop re-entry. The common case (1-3 unrelated comments) skips clustering entirely at zero cost. Same-file concentration is deliberately excluded as a gate signal since it's handled by existing same-file grouping.
- **Cluster detection**: When the gate fires, items are assigned concern categories from a fixed 11-item list, then grouped by category + spatial proximity (same file or directory subtree). Clusters of 2+ items get a `<cluster-brief>` XML block synthesized.
- **Cluster dispatch**: One agent per cluster receives the brief and all related thread details. Individual items dispatch as before.
- **Verify-loop re-entry**: Any re-entry triggers the gate. The cluster analysis step receives just-fixed file context so it can assess whether new feedback relates to the previous cycle's work.
- **Iteration cap**: After 2 fix-verify cycles, remaining issues surface to the user with "recurring pattern" framing instead of looping indefinitely.

**pr-comment-resolver.md** — Extended with cluster mode:

- Mode detection table: `<cluster-brief>` XML block presence activates cluster mode
- Cluster workflow: parse brief → read broader area → assess systemic vs coincidental → holistic or individual fix → return per-thread verdicts + `cluster_assessment`
- New cluster example in `<examples>` section
- Updated description to reflect handling one or more related threads

## Key design decisions

- **Gated, not always-on**: Clustering only runs when signal strength warrants it. False positive gates (gate fires but no cluster found) cost only the analysis step — no extra agent dispatches.
- **Re-entry is the safety net**: If the gate misses a cluster on the first pass (e.g., 3 items below the volume threshold), verify-loop re-entry catches it on the second pass. One extra review cycle is an acceptable tradeoff for keeping the common case fast.
- **Agent decides holistic vs. individual**: The orchestrator detects the cluster and synthesizes the brief, but the agent reads the code and judges whether a structural fix or individual patches are appropriate.
- **No separate cross-cycle mechanism**: Verify-loop re-entry IS the whack-a-mole signal. No concern-category string matching or structural comparison across inference passes — the cluster analysis step handles thematic grouping with just-fixed context.

## Test plan

- Verify `bun test` passes (506 tests, 0 failures) ✅
- Verify `bun run release:validate` passes (48 agents, 41 skills, 1 MCP server) ✅
- Manual verification: all step cross-references updated correctly (steps 3-8→skip, step 8→verify, Targeted Mode steps 6-7, etc.)

---

### DIFF
--- a/plugins/compound-engineering/skills/resolve-pr-feedback/SKILL.md
+++ b/plugins/compound-engineering/skills/resolve-pr-feedback/SKILL.md
@@ -177,7 +177,7 @@ Verdict meanings:
 
 **Batching**: Clusters count as 1 dispatch unit regardless of how many threads they contain. If there are 1-4 dispatch units total (clusters + individual items), dispatch all in parallel. For 5+ dispatch units, batch in groups of 4.
 
-**Conflict avoidance**: If multiple threads reference the same file and are NOT part of a cluster, group them into a single agent dispatch to avoid parallel edit conflicts. The agent handling a multi-thread file receives all threads for that file and addresses them sequentially.
+**Conflict avoidance**: No two dispatch units that touch the same file should run in parallel. Before dispatching, check for file overlaps across all dispatch units (clusters and individual items). If a cluster's file list overlaps with an individual item's file, or with another cluster's files, serialize those units -- dispatch one, wait for it to complete, then dispatch the next. Non-overlapping units can still run in parallel. Within a single dispatch unit handling multiple threads on the same file, the agent addresses them sequentially.
 
 **Sequential fallback**: Platforms that do not support parallel dispatch should run agents sequentially. Dispatch cluster units first (they are higher-leverage), then individual items.
 


tmchow:
@codex review

PR #442, opened 2026-03-29
feat(converters): centralize model field normalization across targets

tmchow:
## Summary

Each target platform expects a different format for the `model` frontmatter field, but converters handled this inconsistently — Qwen had wrong alias mappings (`claude-sonnet` instead of `claude-sonnet-4-6`), OpenClaw and Copilot passed through raw values, and there were two divergent copies of `CLAUDE_FAMILY_ALIASES` across OpenCode and Qwen.

This PR extracts shared model normalization into `src/utils/model.ts` and applies per-target behavior based on external research into each platform's model field format:

| Target | `model: sonnet` output | Rationale |
|--------|------------------------|-----------|
| OpenCode | `anthropic/claude-sonnet-4-6` | Multi-provider, needs provider prefix |
| Qwen | `anthropic/claude-sonnet-4-6` | Multi-provider (supports Anthropic via settings.json) |
| OpenClaw | `anthropic/claude-sonnet-4-6` | Multi-provider, needs provider prefix |
| Droid | `sonnet` (pass-through) | Factory natively resolves bare aliases |
| Copilot | dropped | Supports model but format is display names, not IDs |
| Codex | dropped | Skill frontmatter only supports `name` and `description` |

## Key decisions

- **Droid pass-through**: Factory docs confirm it accepts `sonnet`, `opus`, `haiku` natively and resolves them internally. No normalization needed.
- **Copilot drop**: Copilot supports Claude models but uses display-name format ("Claude Opus 4.5") rather than model IDs. Omitting is safer than guessing the format; docs say "If unset, inherits the default model."
- **Codex drop**: Confirmed via Rust source (`SkillFrontmatter` struct) that skill frontmatter only supports `name` and `description`. Model is set globally in `config.toml`.
- **Qwen fix**: Old code mapped `sonnet` → `claude-sonnet` (wrong). Now uses shared aliases mapping to `claude-sonnet-4-6`.

## Test plan

- New `tests/model-utils.test.ts` covers all shared utility functions including provider prefix routing for `claude-*`, `gpt-*`, `gemini-*`, `qwen-*`, and `o3-*` patterns
- Each converter test file has explicit model handling tests (bare alias resolution, pass-through, drop behavior)
- Full suite: 524 tests pass

---

### DIFF
--- a/docs/solutions/integrations/cross-platform-model-field-normalization-2026-03-29.md
+++ b/docs/solutions/integrations/cross-platform-model-field-normalization-2026-03-29.md
@@ -79,7 +79,7 @@ Each converter now uses the appropriate shared utility:
 | Qwen | Resolve alias + add provider prefix | `anthropic/claude-sonnet-4-6` |
 | OpenClaw | Resolve alias + add provider prefix | `anthropic/claude-sonnet-4-6` |
 | Droid | Pass through as-is | `sonnet` |
-| Copilot | Pass through as-is | `claude-sonnet-4-20250514` |
+| Copilot | Drop entirely | (omitted) |
 | Codex | Drop entirely | (omitted) |
 
 ---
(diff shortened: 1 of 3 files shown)


tmchow:
Re: the Copilot model field discussion — I initially restored pass-through per the review feedback, but on further consideration reverted it.

The issue: Copilot's `model` field expects display names like "Claude Opus 4.5", not Claude model IDs (`claude-sonnet-4-20250514`) or bare aliases (`sonnet`). Unlike Droid/Factory which natively resolves `sonnet`, Copilot has no documented resolution for Claude-specific values. So pass-through would emit a field Copilot can't use — which is arguably worse than omitting it (where the spec says it falls back to the platform default).

The distinction from Droid: Factory explicitly documents that it accepts `sonnet`, `opus`, `haiku` as valid model values. Copilot's model field works with a different value space entirely.

If Copilot model mapping becomes needed later, it would require a dedicated `COPILOT_MODEL_MAP` translating Claude aliases to Copilot display names — similar to how we'd need one for Codex if they ever add per-skill model support.

PR #454, opened 2026-03-30
feat(ce-review): enforce table format, require question tool, fix autofix_class calibration

tmchow:
## Summary

ce:review interactive output was drifting from spec in three ways: findings rendered as freeform prose blocks instead of pipe-delimited tables, the post-review policy question used generic conversational text instead of the platform question tool, and reviewers defaulted nearly everything to `advisory` instead of properly classifying `safe_auto` candidates.

## Changes

**Stage 6 — findings must be tables** (SKILL.md, review-output-template.md)
- Mandate pipe-delimited tables for finding rows specifically, scoped so bullet-list sections (Applied Fixes, Learnings, Coverage) and the `---` verdict separator aren't contradicted
- Add a format verification checkpoint before delivering the report
- Add an anti-pattern section to the output template showing the exact wrong format to avoid

**After Review — name the question tool, fix manual-only prompt** (SKILL.md)
- Explicitly require `AskUserQuestion` / `request_user_input` / `ask_user` with a fallback for environments without one
- Split the policy prompt into two variants: full 3-option when `gated_auto` findings are present, 2-option when only `manual` findings remain (drops the nonsensical "approve gated fixes" option)

**autofix_class calibration** (subagent-template.md, findings-schema.json)
- Replace "set conservatively" with a decision guide and concrete examples for each class
- Align `safe_auto` definition across all three surfaces (SKILL.md, schema, template) as "local and deterministic — no design judgment required"
- Add "do not default to advisory when a concrete fix is obvious"

## Test plan

- `bun test` — 530 pass, 0 fail
- `bun run release:validate` — metadata in sync

---

### DIFF
--- a/plugins/compound-engineering/skills/ce-review/references/review-output-template.md
+++ b/plugins/compound-engineering/skills/ce-review/references/review-output-template.md
@@ -115,7 +115,7 @@ This fails because: no pipe-delimited tables, no severity-grouped `###` headers,
 
 ## Formatting Rules
 
-- **Pipe-delimited markdown tables** -- never ASCII box-drawing characters or horizontal-rule-separated text blocks
+- **Pipe-delimited markdown tables** for findings -- never ASCII box-drawing characters or per-finding horizontal-rule separators between entries (the report-level `---` before the verdict is still required)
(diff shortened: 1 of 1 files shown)


tmchow:
@codex review

PR #455, opened 2026-03-30
Add optimize-loop skill

felixclack:
## Summary

Add an `optimize-loop` skill that runs iterative optimization experiments against a single measurable metric, inspired by [Karpathy's autoresearch](https://github.com/karpathy/autoresearch).

## What it does

The skill interviews the user to establish a goal, metric, measurement command, file scope, and iteration budget. It then:

1. **Profiles** to find where time/resources are actually spent
2. **Loops** N iterations — each makes one change, measures, keeps or reverts
3. **Reports** the total improvement with a summary of what worked

Binary keep/discard with git commits means the branch tip always represents the best-known state.

## Use cases

- Test speed optimization (e.g., `bundle exec rspec packages/intake/spec`)
- Flaky test reduction
- Test coverage improvement
- Query/page load speed
- Memory usage / bundle size

## Battle-tested

Used to optimize Brightside's intake package test suite: **446s to 221s (50.4% faster)** across 17 iterations (15 keeps, 2 discards).

The skill documents 6 proven techniques ranked by typical impact, plus common anti-patterns that don't work (learned by discarding failed experiments during the intake optimization).

## Trigger phrases

"optimize loop", "run optimization", "iterate on", "optimize this", "make this faster", "reduce flakiness", "improve coverage"

### DIFF
--- a/plugins/compound-engineering/skills/optimize-loop/SKILL.md
+++ b/plugins/compound-engineering/skills/optimize-loop/SKILL.md
@@ -1,0 +1,157 @@
+---
+name: optimize-loop
+description: "Iterative optimization loop inspired by Karpathy's autoresearch. Measures a single metric, makes one change per iteration, keeps improvements, discards regressions. Use for test speed, coverage, query performance, flakiness, bundle size, or anything with a clear number."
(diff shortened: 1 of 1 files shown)


felixclack:
Wrong repo — moving to Brightside marketplace plugin instead.

PR #480, opened 2026-04-01
feat(resolve-pr-feedback): add cross-invocation cluster analysis

tmchow:
## Summary

- Replace the dead verify-loop re-entry gate signal with a cross-invocation awareness signal that detects multi-round review patterns on the same PR
- Broaden the `get-pr-comments` script's jq filter to include resolved threads in a new `cross_invocation` envelope — zero additional GraphQL calls
- Add three-mode resolver assessment for cross-invocation clusters: band-aid (redo), correct-but-incomplete (investigate sibling code), sound-and-independent (context only)

## What changed

**`get-pr-comments` script** — The jq filter now also selects resolved threads (`isResolved == true`), sorted by recency, limited to the last 10. Outputs a `cross_invocation` object with `signal` (boolean: resolved + unresolved both exist) and `resolved_threads` array. Existing output keys are unchanged.

**SKILL.md** — The cluster gate table replaces "Verify-loop re-entry" with "Cross-invocation" (`cross_invocation.signal == true`). When the signal fires, resolved threads are included in category assignment and spatial grouping alongside new threads. Cluster briefs gain a `<prior-resolutions>` element. A dispatch boundary rule ensures resolved threads participate in clustering but are never individually re-dispatched. The verify loop is simplified to rely on the cross-invocation signal for re-entry detection.

**pr-comment-resolver agent** — Cluster mode now parses `<prior-resolutions>` and uses a three-mode assessment when prior resolutions are present. The "correct but incomplete" mode is the highest-value case: prior fixes were right but reveal a broader pattern in untouched sibling code. A new example demonstrates this mode.

## Motivation

The verify-loop re-entry gate was architecturally dead — it required new threads to appear between push and verify, but automated reviewers post minutes later while verify runs seconds after push. This left the volume gate (3+ items) as the only working trigger for cluster analysis, missing the exact scenario clustering was designed for: recurring feedback about the same problem class across multiple rounds with 1-2 threads each.

## Test plan

- [x] `bun test` — 534 tests pass
- [x] `bun run release:validate` — metadata in sync
- [ ] Manual: run the skill on a PR with resolved threads to verify the `cross_invocation` envelope appears in script output
- [ ] Manual: run on a PR with both resolved and unresolved threads to verify cluster analysis fires

No additional operational monitoring required — this changes skill orchestration content (markdown instructions and a jq filter), not runtime infrastructure.

### DIFF
--- a/plugins/compound-engineering/skills/resolve-pr-feedback/scripts/get-pr-comments
+++ b/plugins/compound-engineering/skills/resolve-pr-feedback/scripts/get-pr-comments
@@ -37,7 +37,7 @@ query FetchPRFeedback($owner: String!, $repo: String!, $pr: Int!) {
   repository(owner: $owner, name: $repo) {
     pullRequest(number: $pr) {
       author { login }
-      reviewThreads(first: 100) {
+      reviewThreads(first: 50) {
         edges {
           node {
             id
@@ -45,7 +45,7 @@ query FetchPRFeedback($owner: String!, $repo: String!, $pr: Int!) {
             isOutdated
             path
             line
-            comments(first: 50) {
+            comments(first: 10) {
               nodes {
                 id
                 author { login }


tmchow:
> **Include resolved-thread text needed for categorization** ([thread](https://github.com/EveryInc/compound-engineering-plugin/pull/480#discussion_r3024008072))

Addressed in 9c2f5bf1: added `first_comment_body` (the original reviewer feedback) to the resolved thread projection. That's the text the skill needs to assign concern categories during clustering. Full thread history isn't needed — just the initial feedback that describes the concern.

PR #493, opened 2026-04-02
fix(converters): remove invalid tools/infer from Copilot agent frontmatter

datorresb:
- Remove `tools: ["*"]` from agent frontmatter (not a valid Copilot agent attribute)
- Replace deprecated `infer: true` with `user-invocable: true`
- Add ce: → ce- namespace transformation in body content
- Update tests to match new frontmatter output


tmchow:
@codex review

tmchow:
nice job @datorresb !

PR #495, opened 2026-04-03
feat(slack-researcher): add Slack organizational context research agent

tmchow:
## Summary

New `slack-researcher` agent that searches Slack for organizational knowledge -- decisions, constraints, and context that lives in conversations rather than code or docs. Integrates conditionally with `ce:ideate`, `ce:brainstorm`, and `ce:plan` when the Slack MCP plugin is installed and authenticated.

### Access model

The agent uses the authenticated user's own Slack credentials -- identical access to what they'd have searching Slack manually. No elevated permissions, no service accounts.

| Source | Access | Rationale |
|---|---|---|
| Public channels | Searched by default | Visible to entire workspace, safe to persist in shared artifacts |
| Private channels | Searched by default | Same access the user has directly; valuable for project-specific context |
| DMs (1:1 and group) | Never searched | The agent is dispatched automatically behind the scenes by workflow skills -- the user may not realize DMs are being searched. Excluding DMs avoids this awareness gap. |

Output is shaped for documents, not chat: the agent surfaces decisions and constraints, not interpersonal dynamics or off-topic tangents. This is a quality principle, not a security boundary -- the user has the same access either way.

### Conversation reasoning

The agent includes lightweight principles (derived from speech act theory, enterprise chat analysis research, and Slack AI's own architecture) that steer how it interprets conversations:

- Treats threads as atomic units of meaning, extracting the conclusion from final replies
- Reads for commitment arcs (proposal → discussion → convergence), not keywords
- Applies topic-specific supersession (a new message about "database choice" doesn't invalidate an older one about "migration timeline")
- Recognizes that brevity signals agreement and elaboration signals resistance

### Workflow integration

- **ce:ideate** — dispatches in parallel during Phase 1 codebase scan (conditional on `slack_*` tools)
- **ce:brainstorm** — dispatches at the start of Phase 1.1 alongside constraint check and topic scan
- **ce:plan** — dispatches during Phase 1.1 local research. When an origin brainstorm doc contains a Slack context section, it's passed verbatim to the researcher so it can focus on implementation-specific gaps rather than re-searching product context already gathered during brainstorming.

### Token efficiency

Output targets ~500-1500 tokens (down from 2000-3000+ in early iterations) using a compact format: Topic / Summary (1-3 sentences) / Source, plus a cross-cutting analysis section that reasons across findings. Search count scales with dispatch complexity rather than a hard cap.

## Test plan

- [x] Tested with 7 distinct Slack queries across realistic ce:ideate, ce:brainstorm, and ce:plan dispatch scenarios
- [x] Verified conversation reasoning principles improve output quality (commitment arcs, thread atomicity, temporal reasoning)
- [x] Verified token budget compliance (~1000-1400 tokens typical, down from 2000-3000+)
- [x] Verified agent adapts search breadth and output focus based on topic context without caller-awareness
- [x] Verified warm handoff: when prior brainstorm Slack findings are provided, the agent produces complementary implementation-focused context with no redundant searches
- [x] Verified cold start: without prior findings, the agent does a thorough fresh search
- [x] Verified graceful degradation when Slack MCP is unavailable (precondition check returns early)
- [ ] `bun test` — parsing/conversion regression check
- [ ] `bun run release:validate` — plugin manifest consistency

### DIFF
--- a/docs/plans/2026-04-02-001-feat-slack-analyst-agent-plan.md
+++ b/docs/plans/2026-04-02-001-feat-slack-analyst-agent-plan.md
@@ -3,7 +3,7 @@ title: "feat(slack-researcher): Add Slack analyst research agent with workflow i
 type: feat
 status: active
 date: 2026-04-02
-origin: docs/brainstorms/2026-04-02-slack-researcher-agent-requirements.md
+origin: docs/brainstorms/2026-04-02-slack-analyst-agent-requirements.md
 ---
 
 # feat(slack-researcher): Add Slack analyst research agent with workflow integration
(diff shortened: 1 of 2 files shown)


CyberAstronaut101:
Quite literally was just looking at this specific problem +1

PR #498, opened 2026-04-03
fix(openclaw): use sync plugin registration

kesslerio:
## What
- make the generated OpenClaw plugin entrypoint export a synchronous `register(api)` function
- inline converted command bodies into the generated command handlers
- remove async skill preloading from the generated OpenClaw entrypoint

## Why
- OpenClaw ignores plugin registrations that return a promise
- the generated async entrypoint causes the runtime to report: `plugin register returned a promise; async registration is ignored`
- switching to fire-and-forget skill loading fixes the warning but introduces a startup race where commands can return not found before skill files finish loading
- inlining the converted command body removes both issues and keeps registration synchronous

## Tests
- `bun test tests/openclaw-converter.test.ts`

## AI Assistance
- Used: yes
- Testing: targeted automated test (`bun test tests/openclaw-converter.test.ts`) plus local OpenClaw plugin inspection after reinstall

### DIFF
--- a/src/converters/claude-to-openclaw.ts
+++ b/src/converters/claude-to-openclaw.ts
@@ -175,17 +175,16 @@ function buildOpenClawConfig(
 function generateEntryPoint(commands: OpenClawCommandRegistration[]): string {
   const commandRegistrations = commands
     .map((cmd) => {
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks @kesslerio!

PR #518, opened 2026-04-06
fix(ce-plan): accept non-software tasks with adapted planning workflow

### LINKED ISSUE #517: /ce:plan rejects non-software tasks with "This isn't a software planning task"
## Problem

When invoking `/ce:plan` for tasks that aren't strictly software engineering (e.g., "read these PDFs and create a summary," research tasks, content planning), the skill short-circuits with:

> This isn't a software planning task, so I'll skip the ce:plan workflow and just [do the thing] directly.

This defeats the purpose. The ce:plan workflow (structured thinking, step breakdown, user alignment before execution) is valuable for *any* multi-step task, not just code. Users who've adopted ce:plan as their default planning tool get inconsistent behavior depending on whether the skill considers the task "software" enough.

## Expected behavior

ce:plan should apply its structured planning process to any task the user routes through it. The skill description already says "Transform feature descriptions or requirements into structured implementation plans" - but "implementation" doesn't have to mean code.

## Suggested fix

Remove or soften the software-task gate. If the skill detects a non-code task, it should still produce a structured plan, just adapted to the task type (research plan, content plan, etc.) rather than bailing out entirely.

mvanhorn:
## Summary

When users invoke /ce:plan for non-software tasks (research, content planning, trip planning), the skill short-circuits with "This isn't a software planning task" and skips the workflow entirely. There's no explicit gate - the rejection is emergent because the entire SKILL.md frames every concept around code (repo patterns, implementation units, test scenarios, file paths).

Added a task domain classifier and conditional guidance so the structured planning methodology works for any multi-step task. The core approach (problem frame, scope, unit breakdown, acceptance criteria) is domain-agnostic - only the specifics (repo patterns, test scenarios) are software-specific.

## Changes

- Updated description to mention non-software task support
- Added "Task domain flexibility" paragraph after the workflow intro
- Added task domain classifier (Software vs Non-software) in Phase 0.6
- Added skip guidance for repo research agents in Phase 1.1 for non-software tasks
- Adapted Phase 3.5 unit definitions to support deliverables/validation criteria
- Added planning rule exemption for non-software plans (file paths, test scenarios)

Net: +14 lines, -2 lines. Kept additions minimal to avoid inflating token cost.

## Demo

Before/After

## Testing

All 587 tests pass, release metadata validated.

Fixes #517

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/plugins/compound-engineering/skills/ce-plan/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-plan/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ce:plan
-description: "Transform feature descriptions or requirements into structured implementation plans grounded in repo patterns and research. Also deepen existing plans with interactive review of sub-agent findings. Use for plan creation when the user says 'plan this', 'create a plan', 'write a tech plan', 'plan the implementation', 'how should we build', 'what's the approach for', 'break this down', or when a brainstorm/requirements document is ready for technical planning. Use for plan deepening when the user says 'deepen the plan', 'deepen my plan', 'deepening pass', or uses 'deepen' in reference to a plan. Best when requirements are at least roughly defined; for exploratory or ambiguous requests, prefer ce:brainstorm first."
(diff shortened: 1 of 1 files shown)


tmchow:
Closing this in favor of #519 which I like better :). I think it scales better across wider variety of scenarios and keeps the chubbiness out of the main skill as much as possible too.

PR #531, opened 2026-04-08
docs: structured marketplace description with USER LEVEL, PROJECT LEVEL, and BEST SUITED FOR

TroonCataroon:
Adds structured audience sections (USER LEVEL, PROJECT LEVEL, BEST SUITED FOR) to the marketplace description so prospective users immediately see who benefits and which project types fit.


tmchow:
@TroonCataroon Thanks for the contribution.  I'm questioning the value here because it makes the description much much longer and harder to parse especially with where descriptions show up across various surfaces.

PR #573, opened 2026-04-16
fix(install): respect OPENCODE_CONFIG_DIR env var for global config path

cavanaug:
Note: I used an AI agent (upstream-skill) to help draft this, but I (cavanaug) reviewed the write-up and the code. If the tone/format isn't a good fit here, please let me know and I'll adjust the skill.

I noticed the install command and sync registry always resolve the OpenCode global config path to `~/.config/opencode`, ignoring the OPENCODE_CONFIG_DIR environment variable that OpenCode itself honours.

This causes problems when users have a non-standard config location (e.g. on NixOS, in Docker, or with XDG_CONFIG_HOME set), because `bun x compound-engineering install` or `bun x compound-engineering sync` writes to the wrong place while OpenCode reads from wherever OPENCODE_CONFIG_DIR points.

The fix is small: both resolveOutputRoot in `src/commands/install.ts` and the opencode entry in `src/sync/registry.ts` now check OPENCODE_CONFIG_DIR first and fall back to `~/.config/opencode` when it's unset. No behaviour change for users without the variable set.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "compound-plugin",
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -144,7 +144,9 @@ export default defineCommand({
             pluginName: plugin.manifest.name,
             hasExplicitOutput,
           })
-          await handler.write(root, bundle)
+          const writeScope =
+            tool.name === "opencode" && isGlobalOpenCodeConfig ? "global" : handler.defaultScope
+          await handler.write(root, bundle, writeScope)
           console.log(`Installed ${plugin.manifest.name} to ${tool.name} at ${root}`)
         }
 
--- a/src/commands/install.ts
+++ b/src/commands/install.ts
@@ -96,7 +96,7 @@ export default defineCommand({
 
     try {
       const plugin = await loadClaudePlugin(resolvedPlugin.path)
-      const outputRoot = resolveOutputRoot(args.output)
+      const { root: outputRoot, isGlobalOpenCodeConfig } = resolveOutputRoot(args.output)
       const codexHome = resolveTargetHome(args.codexHome, path.join(os.homedir(), ".codex"))
       const piHome = resolveTargetHome(args.piHome, path.join(os.homedir(), ".pi", "agent"))
       const hasExplicitOutput = Boolean(args.output && String(args.output).trim())
@@ -163,6 +163,12 @@ export default defineCommand({
       }
 
       const resolvedScope = validateScope(targetName, target, args.scope ? String(args.scope) : undefined)
+      // For OpenCode, if the output root is the global config dir (default or OPENCODE_CONFIG_DIR),
+      // use "global" scope so writeOpenCodeBundle writes the flat layout regardless of basename.
+      const effectiveScope =
+        targetName === "opencode" && isGlobalOpenCodeConfig && resolvedScope === undefined
+          ? "global"
+          : resolvedScope
 
       const bundle = target.convert(plugin, options)
       if (!bundle) {
@@ -177,9 +183,9 @@ export default defineCommand({
         qwenHome,
         pluginName: plugin.manifest.name,
         hasExplicitOutput,
-        scope: resolvedScope,
+        scope: effectiveScope,
       })
-      await target.write(primaryOutputRoot, bundle, resolvedScope)
+      await target.write(primaryOutputRoot, bundle, effectiveScope)
       console.log(`Installed ${plugin.manifest.name} to ${primaryOutputRoot}`)
 
       const extraTargets = parseExtraTargets(args.also)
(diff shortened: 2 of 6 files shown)


cavanaug:
Pushed a follow-up commit (eab704e) addressing all four Copilot review comments:

**Normalization fixes (registry.ts + install.ts)**

Both files now apply `.trim()` and `expandHome()` before `path.resolve()` on `OPENCODE_CONFIG_DIR`, so tilde paths (`~/opencode`) and whitespace-padded values no longer resolve to the wrong location. In `registry.ts` I extracted a `resolveOpenCodeConfigDir(home)` helper so the normalization is defined once and reused by both `detectPaths` and `resolveOutputRoot` — avoids the ternary duplication that was there before.

**Tests**

- `tests/cli.test.ts`: new test sets `OPENCODE_CONFIG_DIR` to a temp dir, runs `install --to opencode`, and asserts output lands under that dir and not under `~/.config/opencode`.
- `tests/detect-tools.test.ts`: two new tests — one confirms detection when `OPENCODE_CONFIG_DIR` points at an existing dir (with no `~/.config/opencode` present), the other confirms non-detection when the env var points at a non-existent dir (even when `~/.config/opencode` does exist).

All 662 existing tests still pass; the one pre-existing failure (`resolve-base.sh > prefers the PR base remote from gh metadata over origin`) is a `gh`-network timeout unrelated to these changes.

cavanaug:
The Codex bot's concern is valid — pushed a second follow-up (411c420) to fix it.

**Root cause confirmed**

`resolveOpenCodePaths` in `src/targets/opencode.ts` infers global-vs-workspace layout from `path.basename(outputRoot)`. It only recognises `"opencode"` or `".opencode"` as global — any other name (including a custom `OPENCODE_CONFIG_DIR` like `/home/user/myconfig`) falls into the workspace branch and writes `agents/` under a nested `.opencode/` subdirectory, which OpenCode does not treat as its global config tree.

**Fix (411c420)**

- `resolveOutputRoot` in `install.ts` now returns `{ root, isGlobalOpenCodeConfig }` — `true` when the path came from `OPENCODE_CONFIG_DIR` or the default `~/.config/opencode`, `false` when it came from an explicit `--output` flag.
- The install command computes `effectiveScope = "global"` when `targetName === "opencode"` and `isGlobalOpenCodeConfig` is true, then passes it to `target.write()`.
- `writeOpenCodeBundle` forwards `scope` to `resolveOpenCodePaths`, which now treats `scope === "global"` as global layout regardless of basename. The existing basename check is kept as a fallback for callers that don't pass scope.
- The `OPENCODE_CONFIG_DIR` CLI test is updated to assert the flat layout (`<dir>/agents/...`) instead of the nested layout.

Existing behaviour for `--output <custom-path>` (workspace/nested layout) is unchanged — all 16 CLI tests pass.

cavanaug:
> **Treat OPENCODE_CONFIG_DIR as global OpenCode root** — chatgpt-codex-connector[bot]

This was a valuable insight that we missed — thank you. The basename heuristic in `resolveOpenCodePaths` was indeed silently producing the wrong layout for any `OPENCODE_CONFIG_DIR` value whose last path segment isn't literally `opencode` or `.opencode`.

The approach we took (411c420) was to thread an explicit `isGlobalOpenCodeConfig` signal from `resolveOutputRoot` through the install command and into `writeOpenCodeBundle` as `scope="global"`, rather than trying to fix the basename inference itself. The reasoning: the install command is the only place that knows *why* a particular output root was chosen, so it's the right layer to make this determination — and `scope` was already wired through to `write()` but unused for opencode.

That said, we're open to alternative approaches if there's a cleaner way to handle this. A few options we considered but didn't pursue:

- **Always append `/opencode` to `OPENCODE_CONFIG_DIR`** if the basename isn't already `opencode` — preserves the basename heuristic but feels fragile and may not match what OpenCode actually does with the env var.
- **Fix `resolveOpenCodePaths` to accept an explicit `isGlobal: boolean`** instead of `scope` — more explicit but essentially the same as what we did with different naming.
- **Remove the basename heuristic entirely** and always require an explicit scope signal — cleaner long-term but a broader change with more surface area to test.

If there's a preferred approach or an issue with the current one, happy to revise.

cavanaug:
> **Apply OpenCode global scope in `--to all` path** — chatgpt-codex-connector[bot]

Correct — another genuine miss. The `effectiveScope` / `isGlobalOpenCodeConfig` logic from 411c420 was only wired into the single-target branch; the `--to all` loop called `handler.write(root, bundle)` with no scope, leaving the same basename-inference bug in place for that path.

Fixed in c1475a1: the `--to all` loop now passes `scope="global"` for the opencode target when `isGlobalOpenCodeConfig` is true, consistent with the single-target path. All existing tests still pass.

tmchow:
@codex review

tmchow:
Thanks for surfacing this, @cavanaug — the bug is real and your fix was correct. I'm closing this PR in favor of carrying the substance forward on #609, which landed a broader refactor of the OpenCode install path (removed the `sync` command, added per-plugin namespacing to `resolveOpenCodePaths`, etc.).

Your changes to `src/sync/registry.ts` don't have a target anymore — that directory is gone. But the core of your fix (respect `OPENCODE_CONFIG_DIR` in `resolveOutputRoot`, thread `scope="global"` through to the writer so a non-conventional basename still gets the flat layout) applies cleanly to the new code path, and I've ported it there with credit to you via `Co-Authored-By:` in the commit message.

The port is in #609:
- `resolveOutputRoot` in `src/commands/install.ts` now checks `OPENCODE_CONFIG_DIR` with the same fallback to `~/.config/opencode`.
- When the output root is the OpenCode global config tree (env var or XDG default), install passes `scope="global"` to `writeOpenCodeBundle`.
- `resolveOpenCodePaths` uses `scope === "global"` or the conventional basenames to pick the flat layout, so an `OPENCODE_CONFIG_DIR` like `/run/user/1000/opencode-xyz` no longer falls through to the nested `.opencode/` project layout.
- Writer regression test added with a non-conventional basename + `scope="global"`.

Once #609 merges, the fix ships. Thanks again for the careful write-up and the reproducible problem scenarios (NixOS / Docker / custom `XDG_CONFIG_HOME`) — that framing was what let me port it cleanly to the rewritten code.

PR #607, opened 2026-04-20
fix(ce-compound): quote YAML array items starting with reserved indicators

### LINKED ISSUE #606: /ce-compound emits invalid YAML frontmatter when symptoms[] item starts with backtick
## Summary

`/ce-compound` emits invalid YAML frontmatter when a `symptoms:` (or any array-of-strings) item starts with a backtick character. Any YAML 1.2 parser rejects the resulting file with `found character that cannot start any token`. The doc is written to disk successfully, but downstream consumers (`yq`, Python `yaml.safe_load`, validator scripts, other plugins that query frontmatter structurally) silently fail. `ce-learnings-researcher`'s grep-first retrieval still matches on substrings, which masks the problem — the file appears to work until something actually tries to parse it.

## Version

- Plugin: `compound-engineering` v2.68.1 (latest as of filing)
- Harness: Claude Code v2.1.114
- Platform: macOS Tahoe 26.4
- Mode: `/ce-compound` Lightweight — not yet verified in Full mode but the frontmatter-writing step is shared between modes

## Reproduction

1. In a fresh Claude Code session, invoke:

    ```
    Document this: "OrbStack container networking stopped resolving mDNS neighbors on Tahoe after the `sudo dscacheutil -flushcache` workflow — had to restart OrbStack to restore LAN peer discovery, not a DNS cache issue as initially assumed." Run /ce-compound.
    ```

2. Pick Lightweight mode when prompted.

3. Read the resulting file under `docs/solutions/integration-issues/`. Its `symptoms:` array will contain an item that starts with a backtick, e.g.:

    ```yaml
    symptoms:
      - `sudo dscacheutil -flushcache; sudo killall -HUP mDNSResponder` does not restore in-container mDNS
    ```

4. Extract the frontmatter and pipe to `yq`:

    ```bash
    awk '/^---$/{c++; next} c==1{print} c==2{exit}' <file> | yq eval '.' -
    ```

    Expected: parses and prints the frontmatter as a mapping.
    Actual: `Error: bad file '-': yaml: line N: found character that cannot start any token`

## Root cause

YAML 1.2 reserves `` ` `` as an indicator character at scalar start position. The plugin's frontmatter-writing subagent prompt does not include guidance to quote array items that begin with reserved characters. The LLM writes what looks visually correct — markdown-style backticks around a shell command — without recognizing the resulting YAML is invalid.

Issue #602 fixed a similar-class problem for `<skill-name>` tokens in `description:` strings. The fix did not generalize to array-item scalars in `symptoms:`, `applies_when:`, `tags:`, `supersedes:`, etc.

## Suggested fix

In `ce-compound/SKILL.md` Phase 2 (frontmatter-writing), add a YAML-safety rule to the subagent prompt:

> When writing array items in frontmatter fields (`symptoms`, `applies_when`, `tags`, `supersedes`, `superseded_by`, `related_components`), if a string starts with any YAML reserved indicator (`` ` ``, `[`, `*`, `&`, `!`, `|`, `>`, `%`, `@`, `?`) or contains `": "`, wrap it in double quotes. Example: `- "\`sudo cmd\`"` not `` - `sudo cmd` ``.

The same rule would benefit `ce-compound-refresh` Replace/Consolidate actions.

## Workaround

For consumers who need YAML-parsable output today, a local nested AGENTS.md at the output directory (e.g., `docs/solutions/AGENTS.md`) with explicit YAML-safety rules does shape the LLM's output reliably — I've verified this in my own repo. The local fix is a patch; the upstream prompt fix is the durable answer.

## Impact

Users of `docs/solutions/` with downstream tooling (validator scripts, CI hooks, other plugins that structurally query frontmatter) will silently accumulate broken files. The grep-first retrieval in `ce-learnings-researcher` masks the problem during normal use, which makes the failure mode particularly non-obvious.

## Additional context

Discovered via a write test against a repo-owned schema override. Local fix (preventive AGENTS.md rule + gotcha note for future reference) is already applied in my repo. Filing this issue for the upstream plugin so the fix lands at the root.

nathanvale:
## Summary

Fixes #606 — `/ce-compound` emits invalid YAML frontmatter when an array item in `symptoms:`, `applies_when:`, `tags:`, `related_components:` (or any future array-of-strings field) starts with a backtick or other YAML 1.2 reserved indicator. Strict parsers (`yq`, `js-yaml` strict, PyYAML) reject the file; `ce-learnings-researcher`'s grep-first retrieval still matches on substrings, so users silently accumulate unparseable files until a structural consumer fails.

This PR encodes the quoting rule in the support-file contract that every `ce-compound` and `ce-compound-refresh` subagent reads at write time, then reinforces it with one-line pointers at the three `SKILL.md` spots that write or dispatch writes of frontmatter, and a preamble comment on the frontmatter template. A regression test asserts the rule is present in every surface.

Same shape of fix as #603 (which handled the `description:` angle-bracket-token case): codify the rule + add a test.

## Changes

| Surface | Change |
|---|---|
| `references/schema.yaml` | Added a `validation_rules` entry for array-item quoting |
| `references/yaml-schema.md` | New "## YAML Safety Rules" section with indicator list, `": "` trap, and before/after example |
| `assets/resolution-template.md` | Added a preamble comment above each frontmatter block pointing to the rule |
| `ce-compound/SKILL.md` | Added one-line pointer at Phase 2 step 5 (Full mode) and step 3 (Lightweight mode) |
| `ce-compound-refresh/SKILL.md` | Added pointer to the Replace-flow subagent dispatch |
| `tests/compound-support-files.test.ts` | Added 8 regression tests asserting rule presence across all 4 support files and both `SKILL.md` files |

Support-file edits are applied to both skill copies in the same commit to keep `tests/compound-support-files.test.ts`'s existing drift assertions green.

Rule body lives in the on-demand-loaded reference file, not in `SKILL.md` — keeps the load-time token footprint flat per plugin `AGENTS.md` Rationale Discipline.

## Scope Boundaries

- Does **not** auto-repair existing invalid frontmatter in users' repos — this fix is forward-looking. A cleanup utility is a possible follow-up PR.
- Does **not** extend quoting guidance to `description:` or other scalar string fields — #603 already covered that.
- Does **not** audit other skills that write frontmatter (`ce-brainstorm`, `ce-plan`, `ce-release-notes`). Each has its own prompt surface; this PR addresses `ce-compound` and `ce-compound-refresh` only.

## Test plan

- [x] `bun test tests/compound-support-files.test.ts` — 11/11 pass (3 drift + 8 new regression)
- [x] `bun test tests/frontmatter.test.ts` — 192/192 pass (SKILL.md frontmatter validity preserved)
- [x] Mutation test: removing the `validation_rules` entry → 2 expected failures fire (the YAML-safety assertion + the drift test). Rule presence is enforced.
- [x] Full suite: 888/889 pass. The single failure (`tests/resolve-base.sh > resolves against origin/HEAD in a detached shallow checkout`) reproduces on clean `main` in this shallow-clone checkout and is unrelated to these changes.

Plan document: `docs/plans/2026-04-20-001-fix-ce-compound-yaml-safety-plan.md` (included in the commit for traceability).

### DIFF
--- a/docs/plans/2026-04-20-001-fix-ce-compound-yaml-safety-plan.md
+++ b/docs/plans/2026-04-20-001-fix-ce-compound-yaml-safety-plan.md
@@ -1,0 +1,162 @@
+---
+title: "fix(ce-compound): quote YAML array items starting with reserved indicators"
+type: fix
+status: active
+date: 2026-04-20
+---
+
+# fix(ce-compound): quote YAML array items starting with reserved indicators
+
+## Overview
+
+`/ce-compound` emits invalid YAML frontmatter when an array item in any
+frontmatter array-of-strings field (primarily `symptoms:`, `applies_when:`,
+`tags:`, `related_components:`) starts with a backtick (`` ` ``) or other YAML
+1.2 reserved indicator. Strict parsers (`yq`, `js-yaml` strict, PyYAML) reject
+the resulting file. The existing angle-bracket-token guardrail (issue #602,
+fixed in #603) does not generalize to array-item scalars. Teach the
+`ce-compound` and `ce-compound-refresh` skills to quote unsafe array items, and
+add a regression test so future prompt edits do not silently drop the rule.
+
+## Problem Frame
+
+YAML 1.2 reserves `` ` `` as an indicator character at the start of a scalar. When
+the frontmatter-writing subagent (or the Lightweight-mode orchestrator) writes
+markdown-style backtick-wrapped shell commands as array items, the output is
+visually correct markdown but syntactically invalid YAML. Strict parsers reject
+the file; `ce-learnings-researcher`'s grep-first retrieval still matches on
+substrings, which masks the problem — users silently accumulate unparseable
+files. Issue #606 provides the reproduction, impact, and suggested fix.
+
+## Requirements Trace
+
+- R1. New `ce-compound` output (Full and Lightweight modes) produces frontmatter
+  that parses under strict YAML 1.2 even when array items begin with reserved
+  indicator characters.
+- R2. `ce-compound-refresh` Replace-flow subagent output meets the same bar.
+- R3. The YAML-safety rule is captured as a durable contract in the authoritative
+  schema files (not only in prompt prose).
+- R4. A regression test fails if the rule is removed from the prompts or the
+  schema contract, preventing silent drift.
+- R5. Existing broken files already under `docs/solutions/` are out of scope.
+
+## Scope Boundaries
+
+- Do not auto-repair existing invalid frontmatter in users' repos.
+- Do not add a runtime YAML validator step to `ce-compound`.
+- Do not change frontmatter schema fields, enum values, or track rules.
+- Do not extend quoting guidance to `description:` or other scalar fields
+  beyond what #603 already covered.
+
+### Deferred to Separate Tasks
+
+- A one-shot cleanup utility for repairing existing broken files in
+  `docs/solutions/`.
+- Broader YAML-safety audit of other skills that write frontmatter.
+
+## Context & Research
+
+### Relevant Code and Patterns
+
+- `plugins/compound-engineering/skills/ce-compound/SKILL.md` — Phase 2 step 5
+  validates frontmatter; Lightweight mode step 3 writes in a single pass.
+- `plugins/compound-engineering/skills/ce-compound/references/schema.yaml` —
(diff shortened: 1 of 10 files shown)


tmchow:
@nathanvale Thanks for the contribution! 

1. Can you look at the codex feedback?
2. How much testing have you done yourself with this change?

tmchow:
Closing in favor of #613

PR #662, opened 2026-04-23
feat(ce-compound): add headless mode:autofix with depth selector

benwilson:
## Summary

Adds a headless affordance to `/ce-compound` for programmatic callers such as post-merge hooks, CI jobs, and orchestrator pipelines.

- **`mode:autofix`** enables unattended operation. It suppresses the mode-selection prompt, session-history opt-in, Discoverability consent, Phase 2.5 selective-refresh prompt, and the post-run "What's next?" menu.
- **`depth:lightweight|standard|deep`** selects the unattended execution depth and defaults to `lightweight`.
  - `depth:lightweight` -> single-pass documentation, no subagents, no overlap check.
  - `depth:standard` -> Phase 1 parallel research, overlap-update behavior, and Phase 3 specialized reviewers. Session history is off.
  - `depth:deep` -> Standard path plus Session Historian dispatch.

`depth:` is parsed only when `mode:autofix` is present. Invalid autofix depth values halt before subagent dispatch with `ce-compound failed. Reason: unknown depth:<value>. Valid values: lightweight, standard, deep.`

## Motivation

`ce-compound` previously required interactive prompts, which made subprocess invocation unreliable in unattended contexts. A call such as `claude -p "/ce-compound ..."` or `codex exec "Run /ce-compound" --ephemeral` could hang or degrade because there was no user available to answer the documented fallback prompt.

Sibling skills in this plugin already have explicit unattended modes (`ce-compound-refresh` uses `mode:autofix`; `ce-code-review` supports `mode:autofix`, `mode:headless`, and `mode:report-only`; `ce-doc-review` supports `mode:headless`). This PR brings `ce-compound` in line with that pattern.

## Output Contract

The full parser-facing contract lives at `plugins/compound-engineering/skills/ce-compound/references/autofix-output-parse-contract.md`; it is not loaded by the skill at runtime.

### Reliable anchors

Search anywhere in stdout — first-line discipline does not hold; the agent prepends narration before any template block.

**Shape detector substrings:**

| Anchor | Shape | Meaning |
|--------|-------|---------|
| `✓ Documentation complete` | success-complete | A doc was written. The `(mode:autofix)` suffix is unreliable on this line; see best-effort. |
| `✓ No documentation written (mode:autofix)` | no-op | Preconditions unmet; nothing written. |

**No-op block supporting anchors** (present when the no-op detector substring appears):

| Anchor | Meaning |
|--------|---------|
| `Depth:` | Execution path taken (`lightweight\|standard\|deep`). |
| `Reason:` | One-line rationale; may continue on indented lines. |
| `Context considered:` | Bulleted summary of available context. |

### Best-effort fields

Callers MUST NOT hard-fail when these are absent — they are emitted opportunistically and frequently drop in headless runs:

- `✓ Documentation updated` substring (success-updated path is unreachable in `claude -p` — demoted by construction)
- `(mode:autofix)` suffix on success-shape lines (lightweight mode emits its own suffix variants)
- `File created:` / `File updated:` paths
- `Track:` / `Category:` / `Overlap detected:` fields
- `Refresh candidate:` and `Discoverability recommendation:` blocks
- `Phase 1 subagents:` / `Phase 3 specialized reviewers:` blocks

## Testing

- `bun test tests/compound-support-files.test.ts tests/pipeline-review-contract.test.ts` - 51 pass / 0 fail.
- `bun run release:validate` - clean.
- `bun test` overall - 896 pass / 9 fail. The 9 failures are all in `tests/skills/ce-polish-beta-project-type.test.ts` and predate this branch.


plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -30,8 +31,56 @@ These files are the durable contract for the workflow. Read them on-demand at th
...
+| **Interactive** (default) | (none) | User picks Full or Lightweight at the Execution Strategy prompt | All interactive prompts per the documented flow |
+| **Autofix** | `mode:autofix` | Lightweight Mode (single-pass, no subagents, no overlap check) | None — all four interactive prompts skipped |
+| **Autofix-full** | `mode:autofix-full` | Full Mode (Phase 1 parallel research, overlap-update, Phase 3 specialized reviewers) | None — session-history defaults to off; Discoverability becomes a recommendation; "What's next?" suppressed |
+
+**Conflicting mode flags.** If both `mode:autofix` and `mode:autofix-full` appear in arguments, stop before any subagent dispatch and emit: `ce-compound failed. Reason: conflicting mode flags — mode:autofix and mode:autofix-full cannot be combined.` This mirrors the conflict handling in `ce-code-review`.
+
+### Shared autofix rules (both `mode:autofix` and `mode:autofix-full`)
+
+- **Skip all user questions.** Never pause for input — the mode-selection prompt, Discoverability Check consent, and "What's next?" menu are all suppressed. `mode:autofix-full` additionally skips the session-history opt-in (defaults to off — see below).
```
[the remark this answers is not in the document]
benwilson: Fixed in 75b8a728. Added an explicit autofix guard on the Phase 2.5 multi-candidate branch — in both `mode:autofix` and `mode:autofix-full`, the orchestrator now recommends `ce-compound-refresh` via the Follow-up line with the broadest useful scope instead of prompting. Also enumerated this prompt in the shared autofix skip list at line 48 so the pattern is obvious for future readers.

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -30,8 +31,56 @@ These files are the durable contract for the workflow. Read them on-demand at th
 
 When spawning subagents, pass the relevant file contents into the task prompt so they have the contract without needing cross-skill paths.
 
+## Mode Detection
```
tmchow, line 34: This section is the biggest ask.

Compare against `ce-compound-refresh/SKILL.md` (lines 10 through 26: one table row per mode, one Autofix mode rules block) and `ce-code-review/SKILL.md` (lines 35 through 71: same pattern, three modes). Both live inside roughly 20 to 40 lines.

What is here runs to about 40 lines for Mode Detection plus another 55 lines of Autofix output block later. Please collapse to the sibling shape: one Mode table, one Shared autofix rules block, state `Skip all user questions` once rather than restating it in every downstream section.

Please change this before we can consider merging.

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -30,8 +31,56 @@ These files are the durable contract for the workflow. Read them on-demand at th
...
+
+Tokenize `$ARGUMENTS` by whitespace and match recognized mode tokens exactly (so `mode:autofix-full` is a distinct token from `mode:autofix`). Strip the matched token from arguments; treat the remainder as the brief context hint.
+
+| Mode | Token | Execution path | Prompts |
+|------|-------|----------------|---------|
+| **Interactive** (default) | (none) | User picks Full or Lightweight at the Execution Strategy prompt | All interactive prompts per the documented flow |
+| **Autofix** | `mode:autofix` | Lightweight Mode (single-pass, no subagents, no overlap check) | None — all four interactive prompts skipped |
+| **Autofix-full** | `mode:autofix-full` | Full Mode (Phase 1 parallel research, overlap-update, Phase 3 specialized reviewers) | None — session-history defaults to off; Discoverability becomes a recommendation; "What's next?" suppressed |
+
+**Conflicting mode flags.** If both `mode:autofix` and `mode:autofix-full` appear in arguments, stop before any subagent dispatch and emit: `ce-compound failed. Reason: conflicting mode flags — mode:autofix and mode:autofix-full cannot be combined.` This mirrors the conflict handling in `ce-code-review`.
```
tmchow: The tokenization rule above and this whole Conflicting mode flags clause exist because `mode:autofix` is a substring of `mode:autofix-full`. That is a smell to me, so reconsider the two-token design.

Two options that avoid the prefix trap:

(a) Orthogonal tokens: `mode:autofix` with an optional `depth:full` flag. Matches how `ce-code-review` layers `base:` and `plan:` alongside its mode tokens.

(b) Reuse `mode:headless` as the thorough variant and let `mode:autofix` stay the cheap Lightweight variant. Closer to the `ce-code-review` precedent directly.

Either removes the need for the dedicated whole-token rule and removes the second Use-when block below.

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -30,8 +31,56 @@ These files are the durable contract for the workflow. Read them on-demand at th
...
+- **Do not edit instruction files.** The Discoverability Check (AGENTS.md / CLAUDE.md edit) requires user consent. In both autofix modes, surface the gap as a `Discoverability recommendation` line in the output and move on.
+- **Be conservative about writing.** Respect the advisory Preconditions (`problem_solved`, `solution_verified`, `non_trivial`). If the conversation and context hint do not evidence a substantive, verified fix, do **not** write a doc — emit the no-op autofix output (see **Autofix mode output** under Success Output) explaining what was missing.
+- **Always print the full output block.** The success output (or the no-op autofix output when no doc is warranted) is the sole deliverable since no user is present to follow up.
+
+### `mode:autofix` (default headless)
+
+Routes through the **Lightweight Mode** execution path (see the Lightweight Mode section below). Single-pass, no Phase 1 parallel subagents, no overlap check, no Phase 3 specialized reviewers. Overlap drift is accepted and caught later by `ce-compound-refresh`.
+
+Use when:
```
tmchow: These Use-when bullets, and the matching block at line 72 for `mode:autofix-full`, are documentation the runtime agent cannot actually use:

- The frontmatter `description` is what a routing model sees when deciding whether to invoke the skill. The body is not.
- By the time the body loads into context, the invocation decision has already been made. Use-when criteria placed here cannot influence the decision they describe.
- Runtime callers (subprocess invocations, parent skills) invoke by name and do not read the body at runtime. The only reader who actually benefits from Use-when content in the body is a human reading source.

Per `plugins/compound-engineering/AGENTS.md`:

> Every line in SKILL.md loads on every invocation. Include rationale only when it changes what the agent does at runtime.

If these criteria genuinely should guide invocation, put them in the frontmatter `description`. If they are for script authors or PR reviewers, they belong in the PR description or README. Either way, please remove from the SKILL.md body.

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -30,8 +31,56 @@ These files are the durable contract for the workflow. Read them on-demand at th
...
+### `mode:autofix-full` (opt-in thorough headless)
+
+Routes through the **Full Mode** execution path — Phase 1 parallel research subagents, overlap-update behavior, Phase 3 specialized reviewers — with the interactive prompts guarded:
+
+- **Mode-selection prompt** → skipped (Full is already chosen by the caller).
+- **Session-history opt-in** → defaults to **off**. The opt-in is interactive-only; autofix cannot decide between sessions-yes and sessions-no without a user, so the conservative default is off (no Session Historian dispatch). A caller that wants session history included must invoke the skill interactively or run session analysis separately and pass the findings as context.
+- **Discoverability Check consent** → surfaced as a recommendation in the output, per the shared rule above.
+- **"What's next?" menu** → skipped, per the shared rule above.
+
+Phase 1 research subagents (Context Analyzer, Solution Extractor, Related Docs Finder) and Phase 3 specialized reviewers (`ce-performance-oracle`, `ce-security-sentinel`, etc.) run as usual in `mode:autofix-full` — they return text data to the orchestrator and do not themselves prompt the user.
```
tmchow: This paragraph is redundant with Mode Detection, not strictly unread at runtime.

The Mode Detection section already names the four prompts guarded in `mode:autofix-full`: mode-selection, session-history opt-in, Discoverability consent, and "What's next?". Phase 1 and Phase 3 subagent dispatch are not in that guarded list, so the agent already knows they run unchanged. The `run as usual` half of this sentence adds nothing on top of that.

The `they return text data to the orchestrator and do not themselves prompt the user` half is a safety justification for the design choice, not a directive the agent acts on. Without it the agent still dispatches correctly because Full Mode says to.

Please cut.

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -462,6 +515,104 @@ What's next?
 
 **After displaying the success output, present the "What's next?" options using the platform's blocking question tool:** `AskUserQuestion` in Claude Code (call `ToolSearch` with `select:AskUserQuestion` first if its schema isn't loaded), `request_user_input` in Codex, `ask_user` in Gemini, `ask_user` in Pi (requires the `pi-ask-user` extension). Fall back to numbered options in chat only when no blocking tool exists in the harness or the call errors (e.g., Codex edit modes) — not because a schema load is required. Never silently skip the question. Do not continue the workflow or end the turn without the user's selection.
 
+**In both autofix modes (`mode:autofix` and `mode:autofix-full`), do NOT present the "What's next?" prompt.** The success output (or the no-op autofix output below) is the sole deliverable. End the turn cleanly after printing it — autofix callers parse the output programmatically and do not respond to interactive menus.
+
+### Autofix mode output
```
tmchow, line 500: Four near-verbatim output templates, roughly 90 lines, loaded on every invocation. AGENTS.md is explicit about this case:

> For skills that orchestrate many tool or agent calls, extract blocks to references/ when they are conditional (only execute under specific conditions) or late-sequence (only needed after many prior calls) and represent a meaningful share of the skill (~20%+).

Two acceptable shapes:

(a) One parameterized template with a mode-labeled header and conditional sections for Phase 1 and Phase 3 blocks. The no-op shape collapses into the same template with a Reason block.

(b) Extract to `references/autofix-output.md` and leave a one-line stub here.

Please pick one and apply before merge.

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -462,6 +515,104 @@ What's next?
...
+
+In both autofix modes the output is the sole deliverable since no user is present to follow up. Print the full block — do not abbreviate, summarize, or skip sections. All output shapes below are machine-parseable: callers detect which shape was emitted by the first line (`✓ Documentation complete (mode:autofix)`, `✓ Documentation complete (mode:autofix-full)`, `✓ Documentation updated (mode:autofix-full)`, or `✓ No documentation written (autofix mode)`) and extract the file path from the `File created:` or `File updated:` block.
```
tmchow: This sentence does two things. Keep the first, drop the second.

The directive telling the agent to print the full block and not abbreviate, summarize, or skip sections is a real runtime rule. Without it the agent will default to summarizing. Keep it.

The rest (naming which first-line strings callers detect, naming the `File created:` and `File updated:` fields as parser anchors) is one level below what the agent needs to produce correct output. The four template blocks immediately below already show exactly what to write verbatim. The parse-contract detail is parser-author documentation, useful for someone writing a downstream script, not for the agent writing the output.

Keep the print-the-full-block directive. Move the parsing-contract detail to a caller-facing doc (README, PR description, or a `references/` file) and delete from SKILL.md.

docs/solutions/skill-design/compound-skill-improvements.md
```diff
@@ -0,0 +1,110 @@
+---
```
tmchow: Please drop this doc, or shrink it significantly.

`docs/solutions/` is for documented solutions to past problems, the institutional memory that compounds. This is a design rationale for a change that is still in review. `ce-compound-refresh` added `mode:autofix` without a parallel learning doc and I don't want this to become the new expectation.

If there is a reusable pattern worth capturing, the anti-pattern table at the end is the only part that meets the bar, and it belongs in `docs/solutions/skill-design/` as a much shorter pattern doc after merge, not a PR-shaped rationale before merge.

One more specific issue: the doc leaks commit history (for example `originally introduced as mode:autonomous, later renamed to mode:autofix`). That belongs in commit messages, not a learning doc.

tmchow (changes requested):
Thanks for putting this together. The goal is right and I want to land this, but the execution is roughly five times the size of the sibling pattern it claims to follow, and it crosses a few of the plugin's own authoring rules in `plugins/compound-engineering/AGENTS.md`. I've flagged the specifics inline.

Headline asks:

1. Collapse Mode Detection to the shape used by `ce-compound-refresh` and `ce-code-review`. One table, one rules block, state each rule once.
2. Reconsider the two-token design. Orthogonal tokens (`mode:autofix` plus `depth:full`) or reusing `mode:headless` would sidestep the prefix tokenization rule and remove the parallel Use-when blocks.
3. Unify the four Autofix output templates into one parameterized block, or extract them to `references/`. Four near-duplicate blocks loaded on every invocation is exactly the case the AGENTS.md extraction rule was written for.
4. Cut runtime-neutral rationale: the Use-when bullets, the subagent-safety paragraph, the `Note:` inside the output template, and the caller-parsing instructions.
5. Drop or radically shrink `docs/solutions/skill-design/compound-skill-improvements.md`. `ce-compound-refresh` did not ship a parallel learning doc when it added `mode:autofix` and I don't want the precedent.

Happy to iterate. Once the size comes down I expect the rest of the review to move quickly.

tmchow (changes requested):
added comments.

plugins/compound-engineering/skills/ce-compound/SKILL.md
benwilson: Considering going with option (a), orthogonal tokens. The new grammar would be `mode:autofix` + optional `depth:<preset>`:

- `mode:autofix` alone → `depth:lightweight` (default)
- `mode:autofix depth:full` → Full execution path, session history off
- `mode:autofix depth:thorough` → Full execution path, session history on

Three presets rather than the binary `depth:full` flag you proposed.

The reason for the third value: inside Full, the session-history opt-in is a second interactive branch that would otherwise block a headless caller, with no objectively safe default (off is conservative but arbitrary). A binary `depth:full` flag defaults sessions off; callers who want sessions-on have no headless path at all. Three presets let each interactive decision inside the skill map to a caller-overridable `depth:` value, so no decision silently defaults where a user would have been asked.

Happy to collapse to binary `depth:full` (two variants, matching your proposal literally) if you'd rather defer `depth:thorough` until a concrete caller asks for sessions-on headless.

Either way, the orthogonal grammar removes the whole-token tokenization rule, the `Conflicting mode flags` clause, and the parallel Use-when blocks — addresses the tail of this comment plus the separate notes on lines 57 and 70.

plugins/compound-engineering/skills/ce-compound/SKILL.md
benwilson: Done in 19885940. Mode Detection collapsed to 23 lines — one two-row table plus a `depth:<preset>` bullet list, an invalid-depth halt sentence, and a four-bullet Autofix mode rules block. Matches the `ce-compound-refresh` shape (lines 10–26) and aligns with `ce-code-review`'s pattern. The two per-variant subsections (`### mode:autofix` and `### mode:autofix-full`) are gone; their rationale was runtime-neutral and has been cut per your line-57/70 asks.

plugins/compound-engineering/skills/ce-compound/SKILL.md
benwilson: Done in 19885940. Both per-variant Use-when blocks (this one and the matching block on line 72 for `mode:autofix-full`) are removed — the entire `### mode:autofix` and `### mode:autofix-full` subsections went with them when Mode Detection collapsed to the sibling shape. Routing-model guidance lives in the frontmatter `description` only.

plugins/compound-engineering/skills/ce-compound/SKILL.md
benwilson: Done in 19885940. Subagent-safety paragraph removed. Mode Detection now states the behavior once (the Autofix mode rules block lists which prompts are suppressed); the Phase 1/Phase 3 redundancy is gone.

plugins/compound-engineering/skills/ce-compound/SKILL.md
benwilson: Done in 19885940. The four templates collapsed into one parameterized success template plus one short no-op template — roughly 60 lines combined, down from ~90. The consolidated block sits well under the ~20% extraction threshold (630-line SKILL.md × 20% = 126 lines), so it stays in-body. Conditional `[If depth:full|thorough]` blocks cover the Phase 1 / Phase 3 / overlap-update variations inside the same template.

plugins/compound-engineering/skills/ce-compound/SKILL.md
benwilson: Done in 19885940. The "print the full block, do not abbreviate" directive survives as a bullet in the Autofix mode rules block (Mode Detection). The parser-contract paragraph (first-line strings, `File created:` / `File updated:` anchors) is removed from SKILL.md and moved to `plugins/compound-engineering/skills/ce-compound/references/autofix-output-parse-contract.md` — not referenced from SKILL.md, so it stays out of the runtime context. The PR body also carries the contract so downstream callers can find it without loading the reference file.

docs/solutions/skill-design/compound-skill-improvements.md
benwilson: Done in 19885940. `docs/solutions/skill-design/compound-skill-improvements.md` is deleted outright (no shrunken successor). Matches `ce-compound-refresh`'s autofix rollout, which shipped without a parallel learning doc. If the anti-pattern table turns out to be reusable after this lands, a proper pattern doc can follow in a separate PR.

benwilson:
Thanks for the review @tmchow — all five asks addressed in 19885940:

1. **Mode Detection collapsed** (line 34) to 23 lines: one two-row table + `depth:<preset>` bullets + invalid-depth halt + four-bullet Autofix mode rules. Matches the `ce-compound-refresh` shape.
2. **Token grammar reworked** (line 44) to orthogonal `mode:autofix` + optional `depth:lightweight|full|thorough`. No prefix trap, no tokenization rule, no conflicting-mode-flags clause.
3. **Output templates consolidated** (line 520) to one parameterized success template + one short no-op template (~60 lines combined, stays in-body — 630 × 20% = 126-line extraction threshold).
4. **Runtime-neutral rationale cut** (lines 57, 70, 522, 544): per-variant Use-when bullets gone with their parent subsections, subagent-safety paragraph gone, `Note:` inside the output template gone, parser-contract paragraph moved to `references/autofix-output-parse-contract.md` (not loaded at runtime; contract also appears in the refreshed PR body for caller authors). "Print the full block" directive survives in the Autofix mode rules block.
5. **Learning doc deleted** — `compound-skill-improvements.md` is gone with no shrunken successor.

Net: -209 / +91 lines in the commit; SKILL.md dropped from 686 to 630. `bun test` 896/9 (same 9 pre-existing `ce-polish-beta-project-type` failures); `bun run release:validate` clean.

**One pre-authorized divergence from your line-44 suggestion** (follow-up to #discussion_r3140321365): went with three depth presets (`lightweight|full|thorough`) rather than a binary `depth:full`. Rationale: `depth:thorough` is the only way to expose the session-history opt-in to a headless caller without prompting. If you'd rather this collapse to binary, say the word and I'll fold `thorough` into `full` (or drop it entirely). Not planning to defend the three-preset choice past one round.

PR body refreshed to match the new grammar and carries the output contract table.

plugins/compound-engineering/skills/ce-compound/SKILL.md
benwilson: Landed in 19885940. Went with option (a) — orthogonal tokens:

- `mode:autofix` alone → `depth:lightweight` (default)
- `mode:autofix depth:full` → Full path, session history off (replaces former `mode:autofix-full`)
- `mode:autofix depth:thorough` → Full path, session history on (new capability — the only way to opt into Session Historian from a headless caller without prompting)

The prefix trap is gone: no whole-token tokenization rule, no "Conflicting mode flags" clause. Invalid `depth:` values halt with `ce-compound failed. Reason: unknown depth:<value>. Valid values: lightweight, full, thorough.`

One divergence from your binary `depth:full` proposal: three presets instead of two. `depth:thorough` exists specifically to carry the session-history opt-in — without it, headless callers can't reach Session Historian at all, since the prompt is interactive-only. If you'd rather collapse to binary (drop `thorough`, default `depth:full` to session-history-off, document that headless callers pass curated context for session history), I'll fold it. Happy to do that in a follow-up commit rather than defend this past one round.

Mode Detection is 23 lines, output templates consolidated to ~60, learning doc deleted, and the parser contract moved to `references/autofix-output-parse-contract.md` (not loaded at runtime; also surfaced in the refreshed PR body).

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -30,8 +31,38 @@ These files are the durable contract for the workflow. Read them on-demand at th
...
+
+- `depth:lightweight` (default) — Lightweight Mode: single-pass, no subagents, no overlap check.
+- `depth:full` — Full Mode: Phase 1 parallel research, overlap-update, Phase 3 specialized reviewers. Session history off.
+- `depth:thorough` — Full Mode + session history on.
+
+Invalid `depth:` values halt before subagent dispatch and emit `ce-compound failed. Reason: unknown depth:<value>. Valid values: lightweight, full, thorough.`
+
+### Autofix mode rules
+
+- **Skip all user questions.** Never pause — the mode-selection prompt, Phase 2.5 selective-refresh prompt (multi-candidate case), session-history opt-in, Discoverability Check consent, and "What's next?" menu are all suppressed.
```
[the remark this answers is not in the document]
benwilson: Fixed in a8cf13a4.

You were right — the Mode Detection suppression alone wasn't strong enough to override the "present the options ... do not end the turn without the user's selection" imperative sitting at the prompt site. A headless `claude -p --permission-mode bypassPermissions "/ce-compound mode:autofix depth:thorough ..."` run before the fix showed the agent emitting the full numbered menu with a confused inline disclaimer ("Autofix mode was specified, so I'm stopping here rather than prompting"), which is exactly the conflicting-requirements case you flagged.

Added a seven-word guard at the prompt site: *"In autofix mode, skip this step."* Same test run after the fix returns no "What's next?" header, no options, no prompt — clean termination.

Context: my earlier U2 refactor (commit 19885940) deleted a longer version of this guard that used to sit below the prompt instruction, judging it redundant with the Autofix mode rules block. That was too aggressive — the rule is load-bearing at the point of action even when stated in Mode Detection.

plugins/compound-engineering/skills/ce-compound/SKILL.md
```diff
@@ -30,8 +31,38 @@ These files are the durable contract for the workflow. Read them on-demand at th
 
 When spawning subagents, pass the relevant file contents into the task prompt so they have the contract without needing cross-skill paths.
 
+## Mode Detection
+
+Parse `$ARGUMENTS` for `mode:autofix` and an optional `depth:<value>` token. Strip matched tokens; treat the remainder as the brief context hint.
```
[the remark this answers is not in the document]
benwilson: Fixed in 3d032e1d. Two minimal edits:

1. Parse instruction now conditional: `Parse $ARGUMENTS for mode:autofix. If present, also parse an optional depth:<value> token. Strip matched tokens; the remainder is the context hint.` — depth: is only parsed when autofix is present.

2. Halt rule scoped: `In autofix mode, invalid depth: values halt before subagent dispatch and emit ...` — invalid-depth error no longer fires outside autofix.

Verified headless: `/ce-compound depth:bogus` (no `mode:autofix`) now enters Interactive mode instead of emitting the halt envelope. The autofix-path halt still fires correctly on `/ce-compound mode:autofix depth:bogus`.

Net change: 2 insertions / 2 deletions — zero line growth, since both fixes fit on the existing single lines.

benwilson:
Follow-ups since 19885940:

- `6f409f8b` — dispatch Session Historian on `depth:thorough` in autofix. Caught by headless test: Phase 1 step 4 header still said "only if the user opted in," so the autofix dispatch branch never fired even though Mode Detection promised it would.
- `77d9de66` — tighten the Session Historian dispatch rule per AGENTS.md rationale discipline. Collapses parenthetical restatements 6f409f8b had introduced.
- `a8cf13a4` — guard the "What's next?" prompt at the prompt site. Codex P1 `#discussion_r3140572986`. Mode Detection suppression alone was too weak to override the direct "present the options" imperative nearby.
- `3d032e1d` — scope `depth:` parsing and validation to autofix mode. Codex P2 `#discussion_r3140572989`. Interactive calls whose context happened to contain `depth:...` no longer hit the invalid-depth halt.

Known drift worth flagging before re-review: in headless `claude -p --permission-mode bypassPermissions "/ce-compound mode:autofix depth:full|thorough ..."` runs, agents reliably emit `✓ Documentation complete` and `File created:` / `File updated:` anchors, but frequently drop the `(mode:autofix)` suffix, `Depth:`, `Track:`/`Category:`, and the structured `Phase 1 subagents:` / `Phase 3 specialized reviewers:` blocks — synthesizing their own summary with improvised section names instead. I tested six SKILL.md-text interventions (stronger imperatives, extraction to `references/`, explicit tool-call directives, structural marker changes) and none produced reliable contract compliance. Looks like a pattern that doesn't yield to modest text changes — a reliable fix would likely need subagent dispatch for the output step or a structured output format (YAML/JSON), both beyond this PR's review scope. Downstream parsers keying on the reliable anchors (first-line `✓` string + `File created:` / `File updated:`) will work; those depending on the unreliable anchors won't. Flagging so it isn't a surprise if you test headlessly — happy to address in a follow-up PR if it matters.

tmchow (changes requested):
@benwilson I updated the PR description to be better as your agent was weirdly making it into a changelog which wasn't necessary.

I'd like to change the mode values to `lightweight`, `standard` and `deep` to match what we do in other skills (vs `lightweight`, `full`, and `thorough`)

tmchow:
@benwilson reconcile teh PR description to ensure it's accurate given the changes please

benwilson:
Done — PR description reconciled with the current branch. It now uses `depth:lightweight|standard|deep`, updates the invalid-depth error text, and updates the output-contract table.

tmchow (changes requested):
On the headless drift you flagged in https://github.com/EveryInc/compound-engineering-plugin/pull/662#issuecomment-4317095410 — please address before merge rather than deferring. The current `references/autofix-output-parse-contract.md` promises stability for fields you've already confirmed don't hold (`(mode:autofix)` suffix, `Depth:`, `Track:`, `Category:`, `Phase 1 subagents:` / `Phase 3 specialized reviewers:` blocks). Shipping a contract that misleads downstream parsers is worse than shipping a smaller, accurate one — post-merge hooks and CI consumers will key off documented anchors and silently break when they drop.

1. Trim the contract in `references/autofix-output-parse-contract.md` to what actually holds reliably — the first-line `✓ Documentation complete|updated|No documentation written` string and `File created:` / `File updated:` paths. Move every other anchor under a clearly labeled **Best-effort fields (may be dropped in headless runs)** section, with a one-line note that callers must not hard-fail when these are missing.
2. Update the Output contract table in the PR description to match.

Not asking you to land the structural fix here, that's a separate PR can tackle later. Just don't let the doc overstate what the runtime delivers today.

benwilson:
## Headless reliability sweep — empirical findings (PR 662)

Following [#pullrequestreview-4231940863](https://github.com/EveryInc/compound-engineering-plugin/pull/662#pullrequestreview-4231940863), I ran a formal sweep of every parser-anchor field in `references/autofix-output-parse-contract.md` (and the SKILL.md template anchors not currently in the contract) so the reliable/best-effort split is data-driven rather than asserted.

**Bench**

- Branch: `feat/ce-compound-autofix` @ `7c8156d4`
- CWD: `/Users/bwilson/Documents/Source/compound-engineering-plugin` (repo root)
- `docs/solutions/` entry count at SHA: 26
- `claude` CLI: 2.1.129
- Sweep elapsed: 13 minutes (18 total runs, all `exit 0`)

**Methodology**

For each cell `{intended-shape} × {depth}`, I ran 3 invocations of:

```
claude -p --permission-mode bypassPermissions "/ce-compound mode:autofix depth:<preset> <prompt>"
```

The `success-updated` shape was skipped by construction — overlap-update needs the Solution Extractor to produce a "new fix" alongside Related Docs Finder finding "high overlap," and the Solution Extractor reads conversation history, which `claude -p` doesn't have. `Overlap detected:` and `File updated:` are demoted to best-effort by construction.

Two prompts:

- **success-complete:** detailed Jest+Redis fix description with named file, root cause, verification evidence (50 reruns).
- **no-op:** vague "Something seems off with how the build is sometimes slow. Looking into it."

After each run, untracked `docs/solutions/` files were cleaned so subsequent runs started from the same bench state.

**Intended-vs-actual shape map**

| Cell | Run 1 | Run 2 | Run 3 |
|------|-------|-------|-------|
| **complete, lightweight** | complete | complete | other (free-form refusal) |
| **complete, standard** | noop | noop | noop |
| **complete, deep** | noop | noop | noop |
| **noop, lightweight** | other (free-form refusal) | other | other |
| **noop, standard** | noop | other | other |
| **noop, deep** | noop | other | noop |

The standard/deep "complete" cells all flipped to no-op because Phase 1 verification caught that the claimed fix file (`tests/setup/redis-pool.ts`) doesn't exist in this checkout — the agent correctly refused to fabricate a learning doc. Lightweight mode skips that verification step and is more permissive.

**First-line discipline (literal `head -1`)**

Across all 18 runs, the documented `✓ ...` first-line string is the literal first non-blank line in **0/18** runs. The agent always emits narration before the template block. Callers using `head -1` will fail every time.

**`✓` line suffix observed when the shape was emitted**

| Run | Suffix |
|-----|--------|
| complete-lightweight-1 | `(lightweight mode)` |
| complete-lightweight-2 | `(lightweight + autofix mode)` |
| complete-lightweight-3 | (no `✓` line — free-form refusal) |
| complete-standard-1..3 | `(mode:autofix)` *(but on the `✓ No documentation written` line, not `complete`)* |
| complete-deep-1..3 | `(mode:autofix)` *(same as above)* |
| noop-lightweight-1..3 | (no `✓` line — free-form refusal) |
| noop-standard-1 | `(mode:autofix)` |
| noop-standard-2..3 | (no `✓` line — free-form refusal) |
| noop-deep-1, noop-deep-3 | `(mode:autofix)` |
| noop-deep-2 | (no `✓` line — free-form refusal) |

The lightweight mode emits its own suffix variants (`(lightweight mode)` / `(lightweight + autofix mode)`), never `(mode:autofix)`. The lightweight template at SKILL.md:385 (`(lightweight mode)` was added in #528, predates this PR) wasn't updated when this PR added the autofix grammar at SKILL.md:505 promising `(mode:autofix)`. Autofix-lightweight runs fall through to the older lightweight template — a gap this PR introduced by promising a suffix the lightweight path doesn't produce.

**Anchor presence (anywhere in stdout body)**

| Anchor | complete-lw | complete-std | complete-deep | noop-lw | noop-std | noop-deep |
|--------|-------------|--------------|---------------|---------|----------|-----------|
| `✓ Documentation complete` | 2/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `✓ Documentation updated` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `✓ No documentation written` | 0/3 | 3/3 | 3/3 | 0/3 | 1/3 | 2/3 |
| `(mode:autofix)` on `✓` line | 0/3 | 3/3 | 3/3 | 0/3 | 1/3 | 2/3 |
| `^Depth:` | 0/3 | 3/3 | 3/3 | 0/3 | 1/3 | 2/3 |
| `^File created:` | 2/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^File updated:` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^Track:` | 1/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^Category:` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^Overlap detected:` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^Refresh candidate:` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^Discoverability recommendation:` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^Reason:` | 0/3 | 3/3 | 3/3 | 0/3 | 1/3 | 2/3 |
| `^Context considered:` | 0/3 | 3/3 | 3/3 | 0/3 | 1/3 | 2/3 |
| `^Phase 1 subagents:` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |
| `^Phase 3 specialized reviewers:` | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 | 0/3 |

**Reliability classification**

The data does not support keeping anything as "reliable" under a 100% / zero-drift bar. The pragmatic story:

1. **No first-line anchor is reliable.** Callers must search anywhere in stdout, not at line 1.
2. **Phase 1 / Phase 3 subagent blocks are never emitted in headless mode** (0/18 anywhere). They are template aspiration, not runtime reality.
3. **Lightweight mode has its own incompatible output surface** — different `✓` suffix on success, free-form prose on refusal. The contract that nominally covers all three depths actually only fits standard/deep, and only on the no-op path.
4. **The no-op output is the only template that fires consistently when the shape is taken.** When `✓ No documentation written (mode:autofix)` appears, `Depth:`, `Reason:`, `Context considered:` co-occur 9/9 times. That is the strongest reliability signal in the data.
5. **Success-complete anchors (`File created:`, `Track:`, `Category:`) are inconsistent even within the lightweight cell** that did emit the success-complete shape — `File created:` appeared 2/3, `Track:` 1/3, `Category:` 0/3.

**Resulting contract trim**

Reliable section keeps the shape-detector substrings (search anywhere, not first-line) plus the no-op block's supporting anchors:

- `✓ Documentation complete` — success-complete shape; `(mode:autofix)` suffix on this line is unreliable (see best-effort)
- `✓ No documentation written (mode:autofix)` — no-op shape
- When the no-op substring is present: `Depth:`, `Reason:`, `Context considered:` co-occur 100% (9/9 in measured cells)

Best-effort section: `✓ Documentation updated` (unreachable in `-p`, demoted by construction), `(mode:autofix)` suffix on success lines, `File created:`, `File updated:`, `Track:`, `Category:`, `Overlap detected:`, `Refresh candidate:`, `Discoverability recommendation:`, `Phase 1 subagents:`, `Phase 3 specialized reviewers:`. Callers must not hard-fail when these are missing.

Raw run logs are at `${TMPDIR}/compound-engineering/pr-662-headless-reliability/20260505-155532/runs/` on my machine; the methodology and prompts above are sufficient to reproduce from any checkout of the same SHA.

The structural fix to make the demoted anchors reliable (subagent dispatch for the output step, structured output format, or aligning the lightweight template's suffix with the autofix template) is deferred to a follow-up PR per the original review.

benwilson:
Re: [#pullrequestreview-4231940863](https://github.com/EveryInc/compound-engineering-plugin/pull/662#pullrequestreview-4231940863)

Took the empirical route on this — ran an 18-run headless sweep against the current branch so the demotion list is data-grounded rather than asserted. Full methodology and per-cell tallies in [the empirical findings comment](https://github.com/EveryInc/compound-engineering-plugin/pull/662#issuecomment-4385623871).

The data largely supports your reliable set, with two adjustments:

- **Substring matching, not first-line discipline.** No `✓ ...` anchor was the literal first non-blank line of stdout in any run (0/18). The agent always prepends narration before the template block. Reliable claims are scoped to substring presence anywhere in stdout.
- **Demoted `✓ Documentation updated`, `File created:`, `File updated:` to best-effort.** `✓ Documentation updated` is unreachable from `claude -p` (success-updated needs the Solution Extractor to read conversation history, which `-p` doesn't have) and was 0/18. `File created:` was 2/3 only in `complete-lightweight` runs that actually wrote a doc, 0 elsewhere. `File updated:` is in the same demoted-by-construction bucket as `✓ Documentation updated`.

Trimmed `references/autofix-output-parse-contract.md` and the PR body's Output Contract table accordingly:

- **Reliable:** `✓ Documentation complete` and `✓ No documentation written (mode:autofix)` as shape-detector substrings; when the no-op substring is present, `Depth:`, `Reason:`, `Context considered:` co-occur 100% (9/9 in measured cells).
- **Best-effort:** the demoted anchors above plus `(mode:autofix)` suffix on success lines, `Track:`, `Category:`, `Overlap detected:`, `Refresh candidate:`, `Discoverability recommendation:`, and the `Phase 1 subagents:` / `Phase 3 specialized reviewers:` blocks (0/18 anywhere).

One runtime finding worth flagging beyond the contract trim: **autofix-lightweight invocations emit a different suffix than the autofix template promises.** Successes use `(lightweight mode)` or `(lightweight + autofix mode)`, never `(mode:autofix)`. The lightweight template at SKILL.md:385 (predates this PR — `(lightweight mode)` was added in #528) wasn't updated when this PR added the autofix grammar at SKILL.md:505 promising `(mode:autofix)`. Autofix-lightweight runs fall through to the older lightweight template. Reconciling the two is a follow-up PR scope item.

The structural fix to make best-effort anchors reliable (subagent dispatch for the output step or a structured output format) is held for a follow-up PR per your earlier guidance.

PR #681, opened 2026-04-24
fix(ce-code-review): restate model override at dispatch point

### LINKED ISSUE #679: Reviewers spawn as Opus 4.7 instead of Sonnet as expected
When running under opus 4.6 before 4.7 came out my reviewers would spawn as sonnet 90% of the time.

The skill that launches them says they should be run as mid tier models like sonnet but it seems like since 4.7 that is being followed a lot less than it used to when I analyzed my session history.  As a result I'm also burning through usage when reviewer run as opus 4.7 instead of sonnet.  

Anyone else running into this issue?  

Fwiw, these are the types of answers I get when I ask why the reviewers weren't spawned as sonnet 

> ME: I see you ran the reviewer agents as opus 4.7 instead of sonnet, even though the skill directs use of a mid tier model.  Why did you choose not to explicitly pass sonnet as the model?

> CLAUDE: Straight answer: I missed it. The skill text explicitly says "All other persona sub-agents and CE agents use the platform's mid-tier model... In Claude Code, pass model: "sonnet" in the Agent tool call" and names the three exceptions (correctness, security, adversarial). I was focused on scope, reviewer selection, and brief-writing for 12 parallel agents, and didn't re-check Stage 4's model-tiering instruction before dispatch. I should have passed model: "sonnet" to 9 of the 12 Agent calls (everyone except correctness and adversarial — security wasn't dispatched).

> CLAUDE: No defensible reason. That's a cost miss: 9 reviews on Opus 4.7 that should have been on Sonnet 4.6. Nothing to do about this run retroactively — the reviews are done — but I'll save the rule to memory so it carries forward.

tmchow:
Reviewers were dispatching on the session model (Opus 4.7) instead of the mid-tier Sonnet the skill directs, burning through usage on persona reviewers that don't need frontier reasoning. The root cause was placement: the `model: "sonnet"` imperative lived only in a prose "Model tiering" rationale subsection, separated from the "Spawning" subsection that governs actual dispatch. Under 12-agent parallel dispatch, that upstream paragraph fell out of attention.

The fix restates the imperative inside Spawning alongside the existing `mode` paragraph. The 3 session-model exception reviewers (`ce-correctness-reviewer`, `ce-security-reviewer`, `ce-adversarial-reviewer`) are named inline, and the directive covers all cross-platform dispatch primitives (`Agent` in Claude Code, `spawn_agent` in Codex, `subagent` in Pi). The existing "omit if unknown" fallback is preserved for platforms without a model-override parameter. A contract test locks the imperative inside the Spawning subsection so it cannot drift back out to rationale.

Fixes #679

---

Claude Code-D97757?logo=claude&logoColor=white)

### DIFF
--- a/plugins/compound-engineering/skills/ce-code-review/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-code-review/SKILL.md
@@ -402,7 +402,7 @@ Pass the resulting path list to the `project-standards` persona inside a `<stand
 
 Three reviewers inherit the session model with no override: `ce-correctness-reviewer`, `ce-security-reviewer`, and `ce-adversarial-reviewer`. These perform the highest-stakes analysis — logic bugs, security vulnerabilities, adversarial failure scenarios — and should run at whatever capability level the user has configured. If the user is on Opus, these get Opus.
 
-All other persona sub-agents and CE agents use the platform's mid-tier model to reduce cost and latency. In Claude Code, pass `model: "sonnet"` in the Agent tool call. On other platforms, use the equivalent mid-tier (e.g., `gpt-5.4-mini` in Codex as of April 2026). If the platform has no model override mechanism or the available model names are unknown, omit the model parameter and let agents inherit the default -- a working review on the parent model is better than a broken dispatch from an unrecognized model name.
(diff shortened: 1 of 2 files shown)


npwalker (approved):
worked for me

PR #715, opened 2026-04-28
fix(ce-plan): inline post-generation menu routing so option 1 actually starts /ce-work

### LINKED ISSUE #714: fix(ce-plan): post-generation menu option 1 (Start /ce-work) doesn't actually start ce-work
## Bug

`ce-plan` Phase 5.4 presents a 4-option menu where option 1 is "Start \`/ce-work\` (recommended)". When the user picks option 1, the agent renders the menu, gets the selection, and stops — without invoking the `ce-work` skill. The user has to type `/ce-work` themselves.

## Reproduction

1. Invoke `/ce-plan` for any feature
2. Let the skill run through Phase 0 to 5.3 and write the plan file
3. The post-generation menu appears via `AskUserQuestion`
4. Pick option 1, "Start \`/ce-work\` (Recommended)"
5. Observe: the agent acknowledges the choice in prose ("User picked Start /ce-work. Handing off — invoke /ce-work next") and stops, instead of programmatically invoking the `ce-work` skill

## Root Cause

`plugins/compound-engineering/skills/ce-plan/SKILL.md` Phase 5.4 contains the four-option menu list, the \`AskUserQuestion\` invocation guidance, the completion check, and the pipeline-mode exception. But the **routing for each selection** is delegated to \`references/plan-handoff.md\`:

> Routing each selection, contextual surfacing of residual document-review findings, and the post-HITL resync logic all live in \`references/plan-handoff.md\` — follow it for every branch.

Two failure modes follow:

1. **Agent doesn't load the reference.** The skill body says "Load \`references/plan-handoff.md\` now" inside Phase 5.3.8, but agents can render past that and present the menu without ever loading the reference. Without the routing block, the menu becomes a textual handoff with no associated action.
2. **Agent loads the reference but the language is ambiguous.** \`plan-handoff.md\` says \`**Start /ce-work** -> Call /ce-work with the plan path\`. That doesn't name the platform's skill-invocation primitive (Skill in Claude Code, equivalent on Codex/Gemini/Pi), so even an agent that loads it can read "Call /ce-work" as "tell the user to type /ce-work" rather than "fire the Skill tool now."

The plugin's own \`plugins/compound-engineering/AGENTS.md\` "Conditional and Late-Sequence Extraction" section says to extract content that is *conditional or late-sequence and represents ~20%+ of the skill*. The bare per-option routing is late-sequence but **not conditional** — option 1 always means "invoke ce-work." That always-on subset should not have been extracted; only the genuinely conditional flows (HITL state machine, Issue Creation tracker detection) should live in the reference.

## Proposed Fix

1. Inline a \`### Routing\` subsection in \`SKILL.md\` Phase 5.4 with one explicit action per menu option, using platform-explicit invocation language ("invoke the \`ce-work\` skill via the platform's skill-invocation primitive — Skill in Claude Code, equivalent on Codex/Gemini/Pi — passing the plan path as args. Do not merely announce the handoff in prose; actually fire the invocation.")
2. Mirror the platform-explicit phrasing in \`references/plan-handoff.md\`. Keep the elaborate HITL state machine and Issue Creation flows in the reference where they belong.
3. Add a regression test that fails if any of the four inline routing lines are removed in the future.
4. Document the principle in \`docs/solutions/skill-design/\` so future skill authors don't extract always-on routing again.

## Plan Document

Full plan with implementation units, test scenarios, and verification: I have it locally. Will reference it in the PR.

## Severity

UX. The bug doesn't corrupt anything, but it defeats the point of having an interactive menu. The user-facing symptom is "I clicked the recommended option and nothing happened."

mvanhorn:
Closes #714.

## Summary

`ce-plan` Phase 5.4 presents a 4-option post-generation menu where option 1 is "Start \`/ce-work\` (Recommended)". The per-option routing lived only in \`references/plan-handoff.md\`. Two failure modes followed:

1. An agent that didn't load the reference saw the menu as a textual handoff with no associated action and stopped after the user's selection.
2. An agent that did load the reference saw ambiguous language (\`**Start /ce-work** -> Call /ce-work with the plan path\`) that could be read as "tell the user to type /ce-work" rather than "fire the Skill tool now."

Reproduced live on 2026-04-28: user picked option 1, agent acknowledged the choice in prose, and stopped without invoking \`ce-work\`.

## Changes

- **\`plugins/compound-engineering/skills/ce-plan/SKILL.md\`** — Inline \`Routing\` block in Phase 5.4 with one explicit action per menu option, using platform-explicit invocation language ("Invoke the \`ce-work\` skill via the platform's skill-invocation primitive — \`Skill\` in Claude Code, equivalent on Codex/Gemini/Pi — passing the plan path as the skill argument. Do not merely tell the user to type \`/ce-work\` — fire the invocation now.")
- **\`plugins/compound-engineering/skills/ce-plan/references/plan-handoff.md\`** — Mirror the same platform-explicit phrasing in the routing block so the inline and reference language converge. Elaborate sub-flows (Proof HITL state machine, Issue Creation tracker detection, post-HITL \`ce-doc-review\` resync) remain in the reference where they belong.
- **\`tests/skills/ce-plan-handoff-routing.test.ts\`** — Regression test asserting all four inline routing lines are present and that the \`Start /ce-work\` routing specifically names both the skill-invocation primitive and the plan path.
- **\`docs/solutions/skill-design/post-menu-routing-belongs-inline-2026-04-28.md\`** — Authoring principle: always-on routing for interactive menus belongs inline in SKILL.md; only conditional, multi-step sub-flows belong in references. References the existing \"Skill Design Principles\" rule about load-bearing rules in plugin AGENTS.md.

## Why scope this way

The plugin's \`AGENTS.md\` "Conditional and Late-Sequence Extraction" rule says to extract conditional or late-sequence content. The bare per-option routing is late-sequence but **not conditional** — option 1 always means "invoke ce-work." The same AGENTS.md, in "Skill Design Principles," already says: *"For load-bearing rules (those that MUST fire reliably), put strong language at the top of the relevant phase in SKILL.md, not just in the reference."* This PR applies that principle to Phase 5.4 routing while keeping genuinely conditional flows (HITL, Issue Creation) in the reference.

## Verification

- \`bun test tests/skills/ce-plan-handoff-routing.test.ts\` — 3 pass, 0 fail (10 expect calls)
- \`bun test tests/frontmatter.test.ts\` — passes (sanity)
- \`bun run release:validate\` — clean ("Release metadata is in sync. compound-engineering currently has 51 agents, 35 skills, and 0 MCP servers.")
- No version files modified (\`.claude-plugin/plugin.json\`, \`.cursor-plugin/plugin.json\`, \`.codex-plugin/plugin.json\`, \`.claude-plugin/marketplace.json\`, root \`CHANGELOG.md\` — all untouched per AGENTS.md rules)
- Pre-existing failures in \`tests/skills/ce-polish-beta-project-type.test.ts\` (9 fails) confirmed present on \`upstream/main\` before any of this work; unrelated to this PR

## Test plan

- [x] Regression test passes on the fix
- [x] Regression test fails on the regression (verified during authoring — initial regex bugs caught a real "Open in Proof" missing-routing case and a Start /ce-work language-not-platform-explicit case)
- [x] Full \`bun test\` run shows no new failures vs. \`upstream/main\`
- [x] \`bun run release:validate\` passes
- [x] No release-owned files modified

### DIFF
--- a/plugins/compound-engineering/skills/ce-plan/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-plan/SKILL.md
@@ -884,7 +884,7 @@ When deepening is warranted, read `references/deepening-workflow.md` for confide
 
 ##### 5.3.8–5.4 Document Review, Final Checks, and Post-Generation Options
 
-**Load `references/plan-handoff.md` now.** It contains the full instructions for 5.3.8 (document review), 5.3.9 (final checks and cleanup), and 5.4 (post-generation handoff, including the Proof HITL flow, post-HITL re-review, and Issue Creation branching). Document review is mandatory — do not skip it even if the confidence check already ran.
+**STOP. Load `references/plan-handoff.md` now before continuing.** It carries the full instructions for 5.3.8 (document review), 5.3.9 (final checks and cleanup), and 5.4 (post-generation handoff, including the Proof HITL flow, post-HITL re-review, and Issue Creation branching). **This load is non-optional** — without it, the agent renders the post-generation menu, captures the user's selection, and stops without firing the routed action. Document review at 5.3.8 is also mandatory regardless of whether the confidence check already ran.
 
 After document review and final checks, present this menu using the platform's blocking question tool: `AskUserQuestion` in Claude Code (call `ToolSearch` with `select:AskUserQuestion` first if its schema isn't loaded), `request_user_input` in Codex, `ask_user` in Gemini, `ask_user` in Pi (requires the `pi-ask-user` extension). Fall back to numbered options in chat only when no blocking tool exists in the harness or the call errors (e.g., Codex edit modes) — not because a schema load is required. Never silently skip the question.
 
@@ -896,7 +896,7 @@ After document review and final checks, present this menu using the platform's b
 3. **Open in Proof (web app) — review and comment to iterate with the agent** - Open the doc in Every's Proof editor, iterate with the agent via comments, or copy a link to share with others
 4. **Done for now** - Pause; the plan file is saved and can be resumed later
 
-**Routing — act on the user's selection, do not just announce it.** The action for each option is stated inline below so it cannot be missed. The elaborate sub-flows (Proof HITL state machine, Issue Creation tracker detection, post-HITL resync) live in `references/plan-handoff.md`; load that reference for those flows.
+**Routing.** Act on the user's selection — do not just announce it. Elaborate sub-flows (Proof HITL state machine, Issue Creation tracker detection, post-HITL resync) live in `references/plan-handoff.md`.
 
 - **Start `/ce-work`** — Invoke the `ce-work` skill via the platform's skill-invocation primitive (`Skill` in Claude Code, `Skill` in Codex, the equivalent on Gemini/Pi), passing the plan path as the skill argument. Do not merely tell the user to type `/ce-work` — fire the invocation now so the plan executes in this session.
 - **Create Issue** — Detect the project tracker (`gh` for GitHub, `linear` for Linear) and create the issue from the plan file as described under "Issue Creation" in `references/plan-handoff.md`. After creation, display the issue URL and ask whether to proceed to `/ce-work` via the platform's blocking question tool.
--- a/plugins/compound-engineering/skills/ce-plan/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-plan/SKILL.md
@@ -896,8 +896,15 @@ After document review and final checks, present this menu using the platform's b
 3. **Open in Proof (web app) — review and comment to iterate with the agent** - Open the doc in Every's Proof editor, iterate with the agent via comments, or copy a link to share with others
 4. **Done for now** - Pause; the plan file is saved and can be resumed later
 
-Routing each selection, contextual surfacing of residual document-review findings, and the post-HITL resync logic all live in `references/plan-handoff.md` — follow it for every branch.
+**Routing — act on the user's selection, do not just announce it.** The action for each option is stated inline below so it cannot be missed. The elaborate sub-flows (Proof HITL state machine, Issue Creation tracker detection, post-HITL resync) live in `references/plan-handoff.md`; load that reference for those flows.
 
-**Completion check:** This skill is not complete until the post-generation menu above has been presented and the user has selected an action. If you have written the plan file and have not yet presented the menu, you are not done — go to the load instruction above and continue.
+- **Start `/ce-work`** — Invoke the `ce-work` skill via the platform's skill-invocation primitive (`Skill` in Claude Code, `Skill` in Codex, the equivalent on Gemini/Pi), passing the plan path as the skill argument. Do not merely tell the user to type `/ce-work` — fire the invocation now so the plan executes in this session.
+- **Create Issue** — Detect the project tracker (`gh` for GitHub, `linear` for Linear) and create the issue from the plan file as described under "Issue Creation" in `references/plan-handoff.md`. After creation, display the issue URL and ask whether to proceed to `/ce-work` via the platform's blocking question tool.
+- **Open in Proof (web app) — review and comment to iterate with the agent** — Load the `ce-proof` skill in HITL-review mode with the plan file as `source file`, the plan title as `doc title`, identity `ai:compound-engineering` / `Compound Engineering`, and recommended next step `/ce-work`. Then follow the post-HITL resync logic in `references/plan-handoff.md`, which handles the four `ce-proof` return statuses, re-runs `ce-doc-review` after material edits, and falls back gracefully on upload failure.
+- **Done for now** — Display a brief confirmation that the plan file is saved and end the turn. Do not start follow-up work without an explicit further user prompt.
+
+If the user asks for another document review (either from a contextual prompt about residual findings or via free-form request), load the `ce-doc-review` skill with the plan path for another pass and then return to this menu. For free-text revisions outside the four options, accept the input and loop back to this menu after applying the revision.
+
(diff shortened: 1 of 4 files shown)


tmchow (changes requested):
Superseded by inline comments in the follow-up review.

plugins/compound-engineering/skills/ce-plan/SKILL.md
```diff
@@ -896,8 +896,15 @@ After document review and final checks, present this menu using the platform's b
 3. **Open in Proof (web app) — review and comment to iterate with the agent** - Open the doc in Every's Proof editor, iterate with the agent via comments, or copy a link to share with others
 4. **Done for now** - Pause; the plan file is saved and can be resumed later
 
-Routing each selection, contextual surfacing of residual document-review findings, and the post-HITL resync logic all live in `references/plan-handoff.md` — follow it for every branch.
+**Routing — act on the user's selection, do not just announce it.** The action for each option is stated inline below so it cannot be missed. The elaborate sub-flows (Proof HITL state machine, Issue Creation tracker detection, post-HITL resync) live in `references/plan-handoff.md`; load that reference for those flows.
```
tmchow: Cut the second sentence — "stated inline below so it cannot be missed" is reviewer rationale, not agent instruction. The Rationale Discipline rule in `plugins/compound-engineering/AGENTS.md` says cut what wouldn't change agent behavior; the authoring rationale already lives in the new `docs/solutions/` entry.

```suggestion
**Routing.** Act on the user's selection — do not just announce it. Elaborate sub-flows (Proof HITL state machine, Issue Creation tracker detection, post-HITL resync) live in `references/plan-handoff.md`.
```

tmchow: **Re: SKILL.md:887 (the existing `**Load references/plan-handoff.md now.**` line above this hunk):** strengthen it to STOP-tier while you're here. Its current "do not skip" attaches to document review, not the load itself. `ce-brainstorm/SKILL.md:216` and `ce-ideate/SKILL.md:345` already use the stronger pattern (STOP + non-optional + named failure mode).

Replace line 887 with:

> **STOP. Load `references/plan-handoff.md` now before continuing.** It carries the full instructions for 5.3.8 (document review), 5.3.9 (final checks and cleanup), and 5.4 (post-generation handoff, including the Proof HITL flow, post-HITL re-review, and Issue Creation branching). **This load is non-optional** — without it, the agent renders the post-generation menu, captures the user's selection, and stops without firing the routed action. Document review at 5.3.8 is also mandatory regardless of whether the confidence check already ran.

tmchow (changes requested):
Two inline comments below.

tests/skills/ce-plan-handoff-routing.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      { name: "Create Issue", fragment: "Create Issue" },
+      { name: "Open in Proof", fragment: "Open in Proof" },
+      { name: "Done for now", fragment: "Done for now" },
+    ]
+
+    for (const { name, fragment } of optionFragments) {
+      const escaped = fragment.replace(/[.*+?^${}()|[\]\\`]/g, "\\$&")
+      // Bullet form: `- **...<fragment>...**` followed by separator + action.
+      const inlineRoutingPattern = new RegExp(
+        `^- \\*\\*[^\\n]*${escaped}[^\\n]*\\*\\*\\s*(?:[—\\-]+>?|->)\\s*\\S`,
```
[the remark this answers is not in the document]
mvanhorn: Good catch. Fixed in d008977: the inter-token gaps are now `[ \t]*` instead of `\s*`, and the action-text capture is `[^\n]+` so it has to be same-line. Added a dedicated negative-case test that constructs an empty-action bullet followed by a legitimate one — the regex now refuses it, and the existing same-line shape still passes. The original first test would have silently passed on this regression; the new one fails loudly.

plugins/compound-engineering/skills/ce-plan/SKILL.md
mvanhorn: Applied verbatim in d008977. You're right that the second sentence was authoring rationale; the principle-level explanation already lives in `docs/solutions/skill-design/post-menu-routing-belongs-inline-2026-04-28.md`.

mvanhorn: Applied in d008977 — replaced line 887 with the STOP-tier paragraph you drafted, including the explicit failure-mode callout ("renders the menu, captures the user's selection, and stops without firing the routed action"). This is the better root-cause fix for #714 than my routing block alone — the load instruction itself is now non-optional, and the failure mode is named so an agent that's about to skip the load sees the consequence first.

plugins/compound-engineering/skills/ce-plan/SKILL.md
```diff
@@ -896,8 +896,15 @@ After document review and final checks, present this menu using the platform's b
...
 4. **Done for now** - Pause; the plan file is saved and can be resumed later
 
-Routing each selection, contextual surfacing of residual document-review findings, and the post-HITL resync logic all live in `references/plan-handoff.md` — follow it for every branch.
+**Routing.** Act on the user's selection — do not just announce it. Elaborate sub-flows (Proof HITL state machine, Issue Creation tracker detection, post-HITL resync) live in `references/plan-handoff.md`.
 
-**Completion check:** This skill is not complete until the post-generation menu above has been presented and the user has selected an action. If you have written the plan file and have not yet presented the menu, you are not done — go to the load instruction above and continue.
+- **Start `/ce-work`** — Invoke the `ce-work` skill via the platform's skill-invocation primitive (`Skill` in Claude Code, `Skill` in Codex, the equivalent on Gemini/Pi), passing the plan path as the skill argument. Do not merely tell the user to type `/ce-work` — fire the invocation now so the plan executes in this session.
```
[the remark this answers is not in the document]
tmchow: This is fine. Codex can run skills with `$` or via the `Skill` tool

tmchow (approved):

PR #723, opened 2026-04-30
fix(ce-sessions): 722 ce-compound and ce-sessions permission error

davidpgero:
/ce-compund and ce-sessions have permission error. 

Claude Code's permission checker treats piped commands as separate operations and may require approval for each

Resolves https://github.com/EveryInc/compound-engineering-plugin/issues/722

Issue #722, opened 2026-04-30
/ce-compound Error: Shell command permission check failed for pattern


davidpgero:
After I update it, it runs into a problem when I try to use compound. 

version: `3.3.2`

claude code

```bash
/ce-compound
  ⎿  Error: Shell command permission check failed for pattern "!git rev-parse --path-format=absolute --git-common-dir 2>/dev/null | sed -E 's|/\.git/?$||; s|.*/||'": This Bash command contains multiple operations. The following parts require
     approval: git rev-parse --path-format=absolute --git-common-dir, sed -E 's|/.git/?$||; s|.*/||'
```

davidpgero:
Probably: Claude Code's permission checker treats piped commands as multiple separate operations and blocks the skill from loading with: "This Bash command contains multiple operations. The following parts require approval."

PR #737, opened 2026-05-01
fix(ce-compound): update pre-resolved repo name command

DeiAsPie:
image

The PR is a simple command swap to fix the above issue in claude code

### DIFF
--- a/plugins/compound-engineering/skills/ce-compound/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-compound/SKILL.md
@@ -22,7 +22,7 @@ Captures problem solutions while context is fresh, creating structured documenta
 
 ## Pre-resolved context
 
-**Repo name (pre-resolved):** !`git rev-parse --show-toplevel | xargs basename`
+**Repo name (pre-resolved):** !`git rev-parse --show-toplevel | sed 's|.*/||'`
 
 **Git branch (pre-resolved):** !`git rev-parse --abbrev-ref HEAD 2>/dev/null || true`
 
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks for catching this and putting up a fix, @DeiAsPie! Unfortunately we can't merge this one. but the symptom you saw is real and was actually fixed yesterday in #723.

**What's going on:**

The line you're replacing is exactly the line #723 (commit `8f804669`, merged ~17h ago) introduced *to fix* the permission error you're seeing. The original bug (issue #722) was that the previous code shape — `git rev-parse … | sed -E '…'` — trips Claude Code's permission checker as "multiple operations requiring approval" and fails to load the skill at session start. #723 replaced the pipe-to-sed with bash parameter expansion specifically to avoid that.

The reason you're still hitting it is almost certainly that your installed plugin is older than #723. Reinstalling the plugin (or running with `--plugin-dir` against a fresh checkout of `main`) should resolve it without any code change.

**Why we can't merge as-is, even ignoring the install-staleness angle:**

1. **It re-introduces the original bug.** `git rev-parse --show-toplevel | sed 's|.*/||'` has the same top-level pipe shape that triggers the permission prompt — same failure mode as before #723.
2. **It will fail CI.** #723 added `tests/skill-shell-safety.test.ts`, which asserts no `!` pre-resolution command contains a top-level pipe. The new line trips that check, so the test suite will reject it.
3. **It regresses worktree handling.** The current code uses `--git-common-dir` deliberately so the repo name resolves correctly even from inside a git worktree. `--show-toplevel | sed 's|.*/||'` returns the *worktree* directory's leaf name — e.g. inside `/.../workspaces/myrepo/branch-foo` you get `branch-foo` instead of `myrepo`, which breaks the session-historian dispatch hint downstream.

Going to close this, but please grab the latest `main` (or update to 3.4.0+) and let us know if the error reproduces against the updated code; if it does, that's a separate bug worth filing on top of #722.

PR #745, opened 2026-05-01
fix(refresh): skip _shim-* files in orphan check

jcasimir:
## Problem

`generate-shims.sh` produces `_shim-*.md` files in `agents/review/` as generated artifacts, not synced content. The orphan check in `sync-reviewers.sh` flagged every one of them as `Orphan: _shim-X.md (not in any configured source)` on every sync — ~20 lines of pure noise that made it harder to spot real orphans (e.g., when removing a configured source and wanting to verify cleanup of the files it had been producing).

## Fix

Extends the existing `_template-reviewer.md` skip to a `case` glob that also matches `_shim-*`. One-line change in concept; three lines in the diff because `case` is more readable than chaining `[ ... ] && continue` checks for two patterns.

```diff
   filename=$(basename "$filepath")
-  [ "$filename" = "_template-reviewer.md" ] && continue
+  case "$filename" in
+    _template-reviewer.md|_shim-*) continue ;;
+  esac
```

The convention is now: underscore-prefixed files in `agents/review/` are internal/generated and excluded from orphan reporting. This matches how `_template-reviewer.md` was already treated.

## Discovery context

Found while cleaning up duplicate orchestrator-as-reviewer sources in `reviewer-sources.yaml` (an entry that pulled orchestrator files into both `orchestrators/` and `agents/review/`). Every reviewer sync produced a wall of orphan warnings for the auto-regenerated shim files, drowning the legitimate cleanup signal.

## Test plan

- [ ] Run `bash sync-reviewers.sh <registry> <output-dir>` against a directory containing only `_shim-*.md` files — confirm no orphan warnings emit
- [ ] Run against a directory with a real orphan (a `.md` file from a now-removed source) — confirm the orphan warning still fires


jcasimir:
Wrong target — this fix needs to land on the JumpstartLab fork (where the refresh skill still lives at v2.61.0). EveryInc upstream has moved to v3.4.1 with the refresh skill replaced. Reopening on JumpstartLab/compound-engineering-plugin.

PR #752, opened 2026-05-02
fix(ce-compound, ce-sessions): remove bash parameter expansion from ! backtick 

davidpgero:
Claude Code's permission checker rejects bash parameter expansion operators (${var%pattern}, ${var##pattern}, etc.) as "Contains expansion", causing skill load to fail before the skill body ever runs.

```bash
/ce-compound
  ⎿  Error: Shell command permission check failed for pattern "!common=$(git rev-parse --path-format=absolute --git-common-dir 2>/dev/null); repo="${common%/.git}"; echo "${repo##*/}"": Contains expansion
```

`ce-compound` and `ce-sessions` used ${common%/.git} and ${repo##*/} to derive the repo name from git-common-dir. Both skills now drop the repo-name pre-resolution; the skill body already had a fallback instructing the agent to derive the value at runtime via a Bash tool call.

`ce-work-beta` and `ce-product-pulse` used the same pattern in their config pre-resolution to fall back to the main checkout root when running in a worktree. The worktree fallback is removed; config from the worktree's own --show-toplevel path is used instead. Config is optional, so missing it in a worktree is acceptable.

`AGENTS.md` now documents "Contains expansion" as a fourth rejected pattern and removes the incorrect prior advice to use ${var%/suffix} as a fix for the "Unhandled node type: string" case. The config-reading example is updated to match. Two stale error messages in the test file that suggested parameter expansion as a workaround for pipes and nested-quoted-string rejections are also corrected.

`tests/skill-shell-safety.test.ts` gains a hasParameterExpansion detector, 7 unit tests, and a per-file enforcement check that will catch this class of regression going forward.

After the fix:
```bash
claude --plugin-dir local
❯ /compound-engineering:ce-compound

─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ☐ Mode

Which mode would you like to use for documenting this solution?

❯ 1. Full (Recommended)
     Complete compound workflow. Researches, cross-references, and reviews your solution to produce documentation that compounds your team's knowledge.
  2. Lightweight
     Same documentation, single pass. Faster and uses fewer tokens, but won't detect duplicates or cross-reference existing docs. Best for simple fixes or long sessions nearing context limits.
  3. Type something.
```

Reference: https://github.com/EveryInc/compound-engineering-plugin/pull/723


davidpgero:
@tmchow I tested this version locally, and it works well. 

But maybe there is a better solution for it 🤔

tmchow:
Nice work!

PR #754, opened 2026-05-02
fix(ce-code-review): keep finding numbers stable

### LINKED ISSUE #743: For ce code review: Please make auto fix  numbers sequential
I keep a local failure md file that logs each failure and I have Claude sonnet do the code review flow and I usually say "apply auto fixes and append the rest the rest to the failure log" and have codex or opus fix the big issues in a new isolated session. If all the proposed auto fixes were grouped together then I could just copy the non auto fixable findings without having to copy each one.

hnshah:
## Summary
- Assign ce-code-review finding numbers once after sorting, rather than per severity/route bucket
- Reuse stable finding IDs when residual/actionable findings are repeated after safe autofixes
- Tighten the autofix return contract so downstream users can copy non-auto-fixable work without remapping IDs

Fixes #743.

## Test plan
- `bun test tests/review-skill-contract.test.ts tests/pipeline-review-contract.test.ts`
- `bun run release:validate`
- `git diff --check`


plugins/compound-engineering/skills/ce-code-review/SKILL.md
```diff
@@ -551,6 +551,7 @@ Demotion is intentionally narrow. The conservative scope (testing/maintainabilit
    - residual actionable queue: unresolved `gated_auto` or `manual` findings whose owner is `downstream-resolver`
    - report-only queue: `advisory` findings plus anything owned by `human` or `release`
 9. **Sort.** Order by severity (P0 first) -> anchor (descending) -> file path -> line number.
+9b. **Assign stable finding numbers once.** After sorting, assign monotonically increasing `#` values across the full primary finding set. Do not restart numbering inside each severity table or autofix/routing bucket. If later sections repeat a finding (for example Residual Actionable Work after `safe_auto` fixes are applied), reuse the same stable `#` so users can copy the non-auto-fixable work without remapping IDs.
```
tmchow: **Drop the `9b.` workaround -- renumber, or fold into step 9.**

Adding `9b.` between existing steps 9 and 10 looks like a workaround to avoid renumbering `10. Collect coverage data` and `11. Preserve CE agent artifacts.` But these are private pipeline step numbers, not external references -- there's no `ce-work` analog citing them. The stability discipline `ce-plan` teaches applies to user-facing R/U IDs, not internal step numbers.

Pick one:

- **Preferred:** renumber to `10. **Assign stable finding numbers once.**` and bump the existing 10/11 to 11/12.
- **Even better:** fold into existing step 9. Sorting and ID assignment are conceptually one operation -- the stable `#` *is* the post-sort position. Rewrite step 9 as:

  > `9. **Sort and number.** Order by severity (P0 first) -> anchor (descending) -> file path -> line number, then assign monotonically increasing # values across the full primary finding set in that sorted order.`

The folded version is clearer because it makes the dependency explicit: numbering follows from the sort key, so they belong together.

tmchow: **Name the failure mode in prose.**

Current text ends with *"so users can copy the non-auto-fixable work without remapping IDs."* That's generic. Compare to `ce-plan`'s analogous rule (`plugins/compound-engineering/skills/ce-plan/SKILL.md:401` and the deepening-vector callout in `references/deepening-workflow.md:230,244`), which names a specific consumer ("breaks `ce-work` blocker and verification references"). Suggested replacement:

```suggestion
9b. **Assign stable finding numbers once.** After sorting, assign monotonically increasing `#` values across the full primary finding set. Do not restart numbering inside each severity table or autofix/routing bucket. If later sections repeat a finding (for example Residual Actionable Work after `safe_auto` fixes are applied), reuse the same stable `#` so users -- and downstream skills like `ce-resolve-pr-feedback` -- can reference findings by `#` after the autofix loop rewrites the report. Renumbering after autofix invalidates any prior reference: copied snippets, follow-up prompts citing `#3`, or tickets filed against an earlier render.
```

Naming the concrete failure mode lets the next reader judge edge cases the rule didn't anticipate.

plugins/compound-engineering/skills/ce-code-review/SKILL.md
```diff
@@ -600,7 +601,7 @@ When Stage 5b does not run, the merged finding set from Stage 5 flows through to
...
 
 1. **Header.** Scope, intent, mode, reviewer team with per-conditional justifications.
-2. **Findings.** Rendered as pipe-delimited tables grouped by severity (`### P0 -- Critical`, `### P1 -- High`, `### P2 -- Moderate`, `### P3 -- Low`). Each finding row shows `#`, file, issue, reviewer(s), confidence, and synthesized route. Omit empty severity levels. Never render findings as freeform text blocks or numbered lists.
+2. **Findings.** Rendered as pipe-delimited tables grouped by severity (`### P0 -- Critical`, `### P1 -- High`, `### P2 -- Moderate`, `### P3 -- Low`). Each finding row shows `#`, file, issue, reviewer(s), confidence, and synthesized route. Omit empty severity levels. Never render findings as freeform text blocks or numbered lists. Finding numbers must come from Stage 5's stable numbering and continue sequentially across severity sections; never restart at `1` for a new severity header.
```
tmchow: **Collapse the duplicated rule -- point back to Stage 5 instead of restating.**

This sentence restates the rule already established in step 9b: *"Finding numbers must come from Stage 5's stable numbering and continue sequentially across severity sections; never restart at `1` for a new severity header."*

`ce-plan` keeps its stability rule in **one** canonical paragraph (`SKILL.md:401`) and references it from elsewhere -- so future edits have one home, not three (autofix bullet at L70, Stage 5 step, this Stage 6 rendering bullet). Suggested replacement:

```suggestion
2. **Findings.** Rendered as pipe-delimited tables grouped by severity (`### P0 -- Critical`, `### P1 -- High`, `### P2 -- Moderate`, `### P3 -- Low`). Each finding row shows `#`, file, issue, reviewer(s), confidence, and synthesized route. Omit empty severity levels. Never render findings as freeform text blocks or numbered lists. Finding numbers come from the stable assignment in Stage 5 -- never re-derive them per severity table.
```

plugins/compound-engineering/skills/ce-code-review/SKILL.md
```diff
@@ -67,7 +67,7 @@ Sequence:
...
-- **Emit a compact Residual Actionable Work summary in the autofix return** listing each residual `downstream-resolver` finding with severity, file:line, title, and autofix_class. Include the run-artifact path. Callers read this summary directly without parsing the artifact. When no residuals exist, state `Residual actionable work: none.` explicitly.
+- **Emit a compact Residual Actionable Work summary in the autofix return** listing each residual `downstream-resolver` finding with its stable `#`, severity, file:line, title, and autofix_class. Keep applied `safe_auto` fixes and residual non-auto findings in separate contiguous sections, and reuse the stable finding numbers from Stage 5 rather than renumbering residuals. Include the run-artifact path. Callers read this summary directly without parsing the artifact. When no residuals exist, state `Residual actionable work: none.` explicitly.
```
tmchow: **Minor: split the two unrelated rules into separate sentences.**

The new clause mixes section structure ("keep applied fixes and residuals in separate contiguous sections") with ID reuse ("reuse stable finding numbers") into one sentence. Splitting reads more cleanly:

> *"Structure the summary as two separate contiguous sections: applied `safe_auto` fixes first, then residual non-auto findings. Within the residual section, reuse each finding's stable `#` from Stage 5 -- never renumber."*

plugins/compound-engineering/skills/ce-code-review/references/review-output-template.md
```diff
@@ -117,6 +117,7 @@ This fails because: no pipe-delimited tables, no severity-grouped `###` headers,
 
 - **Pipe-delimited markdown tables** for findings -- never ASCII box-drawing characters or per-finding horizontal-rule separators between entries (the report-level `---` before the verdict is still required)
 - **Severity-grouped sections** -- `### P0 -- Critical`, `### P1 -- High`, `### P2 -- Moderate`, `### P3 -- Low`. Omit empty severity levels.
+- **Stable sequential finding numbers** -- assign finding numbers once after sorting, continue them across severity sections, and reuse those same numbers when findings are repeated in Residual Actionable Work. Do not restart at `1` for each severity or route bucket.
```
tmchow, line 120: **Show the rule in the example tables, don't just declare it as a bullet.**

This template already teaches by example -- good rendering above (~L70-100) and bad rendering (~L100-114). The new rule appears only as a declarative bullet here. LLMs rendering the report lean on the example more than the bullet.

Two concrete updates to the example section above:

1. Update the existing P1 / P2 sample tables so the P2 section starts at `#3` (not `#1`), continuing from the P1 section's two findings -- demonstrating the cross-severity continuation.
2. Add a small Residual Actionable Work mini-table below that re-cites one of those `#` values, demonstrating the post-autofix reuse.

Without an example, the rule is easy to read past.

tests/review-skill-contract.test.ts
```diff
@@ -731,8 +731,24 @@ describe("ce-code-review contract", () => {
...
+      "plugins/compound-engineering/skills/ce-code-review/references/review-output-template.md",
+    )
+
+    expect(content).toContain("Assign stable finding numbers once")
+    expect(content).toContain("Do not restart numbering inside each severity table or autofix/routing bucket")
+    expect(content).toContain("reuse the same stable `#`")
+    expect(content).toContain("never restart at `1` for a new severity header")
+    expect(template).toContain("Stable sequential finding numbers")
+    expect(template).toContain("reuse those same numbers when findings are repeated in Residual Actionable Work")
+  })
```
tmchow, line 773: **Assertions are too loose -- tighten location anchoring and consider one behavioral check.**

Two issues with the current test:

1. `expect(content).toContain("Assign stable finding numbers once")` passes even if someone moves the rule into a stale comment, deletes the surrounding step, or adds contradictory guidance elsewhere. The test confirms the rule is *written*, not that it lives in the right place.
2. It checks prose presence, not behavior. Compare to `ce-plan`'s Phase 6 checklist (`SKILL.md:802`) which asserts structural invariants ("U-IDs are unique within the plan and follow the stability rule").

Two concrete improvements -- either is worth doing; both is better:

**(a) Anchor the prose assertions to their location** by slicing the file around the Stage 5 / Stage 5b headings so the test fails loudly if the rule is moved out of context:

```ts
const stage5 = content.split("### Stage 5b:")[0].split("### Stage 5:")[1]
expect(stage5).toMatch(/Assign stable finding numbers once/)
expect(stage5).toMatch(/reuse the same stable `#`/)
```

**(b) Add at least one behavioral check.** Even if the rendering itself lives in skill prose (LLM-driven), you can add a fixture in `tests/fixtures/` showing a correctly-rendered multi-severity report and assert:

- numbers across `### P0`, `### P1`, `### P2`, `### P3` tables are strictly monotonic (never reset to 1)
- a finding repeated in a Residual Actionable Work section retains the same `#`

A fixture-based assertion exercises a concrete artifact rather than a string in skill prose, and would catch the actual failure mode the rule is trying to prevent.

tmchow (commented):
Thanks for tackling this -- the core idea (assign once, reuse downstream) is right. Comparing against how `ce-plan` and `ce-brainstorm` handle stable IDs (R-IDs, U-IDs, etc.) surfaced a few places this could be tightened. Inline comments below; the priority items are the two on `SKILL.md` line 554 (step renumbering / fold into step 9) and the test loosening on the new contract test. Everything else is polish.

For reference, the durable-ID pattern those skills use lives in `plugins/compound-engineering/skills/ce-plan/SKILL.md:401` and `plugins/compound-engineering/skills/ce-plan/references/deepening-workflow.md:230` -- worth a skim if you haven't already, since the language patterns there transfer well.

tmchow (changes requested):
feedback to address posted

lilis-tokyo (approved):

lilis-tokyo (approved):

lilis-tokyo (approved):

tmchow (approved):

PR #760, opened 2026-05-04
feat(ce-brainstorm,ce-plan): add lean/standard local config for ce-brainstorm and ce-plan

davidpgero:
Sometimes I want to have lighter brainstorming and planning sessions.

So this PR adds optional keys to override the mode.

- Optional `.compound-engineering/config.local.yaml` keys `ce_brainstorm_mode` and `ce_plan_research_mode` to reduce workflow heaviness per repo. 
- Default invalid/missing values to `standard` (so no change)


davidpgero:
Sorry, I don't think it's a good idea.

PR #763, opened 2026-05-04
feat(ce-plan): add on-demand modes for deepening and doc review to reduce default planning overhead

davidpgero:
- Add ce_plan_deepening_mode (auto | on_demand) to control whether plan deepening runs automatically.
- Add ce_plan_doc_review_mode (auto | on_demand) to control whether post-plan ce-doc-review runs automatically.
- Update plan handoff behavior so explicit user requests still run deepening/review regardless of mode.

## Why
- Current defaults can feel too heavy for day-to-day planning.
- Teams want to keep brainstorm/plan quality, but only pay deeper review costs when needed.
- This preserves existing default behavior (auto) while enabling a lighter local workflow via config.

## Config
ce_plan_deepening_mode: on_demand
ce_plan_doc_review_mode: on_demand

### DIFF
--- a/.compound-engineering/config.local.example.yaml
+++ b/.compound-engineering/config.local.example.yaml
@@ -11,6 +11,11 @@
 # work_delegate_model: gpt-5.4   # any valid codex model (omit to use ~/.codex/config.toml default)
 # work_delegate_effort: high     # minimal | low | medium | high | xhigh (omit to use ~/.codex/config.toml default)
 
+# --- Brainstorm / Plan pacing ---
+
+# ce_plan_deepening_mode: on_demand  # auto | on_demand (default: auto)
+# ce_plan_doc_review_mode: on_demand # auto | on_demand (default: auto)
+
 # --- Product pulse ---
 # Settings written by /ce-product-pulse first-run interview. Re-run the skill with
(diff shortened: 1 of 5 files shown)


davidpgero:
@codex review

tmchow:
I have to think about this.. i'm not sure I want to regress the auto deepening vs tightenign the rules to be better balanced at when deepening isn't required. did you explore that?

davidpgero:
> I have to think about this.. i'm not sure I want to regress the auto deepening vs tightenign the rules to be better balanced at when deepening isn't required. did you explore that?

Yes, you're right. I don't like it either 🤔 

The reason I introduced `on_demand` is practical: in real use, automatic deepening sometimes goes deeper than the task actually needs, and the credit cost becomes too high.

In some cases, that extra depth is absolutely the right call. In others, it’s unnecessary overhead.

But I only find out too late :( 

So the intent is not to weaken quality gates, but to separate:
- Necessity detection (is deeper analysis needed?)
- Execution timing (run now vs defer)

I want to regain control over when deepening runs, without losing the quality signal.

I’m going to leave this as a draft for now and sleep on it.

davidpgero:
sorry, yes, it's not a good direction.

PR #812, opened 2026-05-10
fix(ce-code-review): resolve base helper from skill dir

### LINKED ISSUE #811: ce-code-review resolve-base helper path resolves against project root
Howdy folks! Thanks for sharing this plugin and skills with the world. Human here steering my agent to, hopefully, be helpful. 😇 I noticed a small bug here related to #764 for `ce-code-review`. I thought my clanker did a better job articulating this below so I'll defer the details:

## Problem

The `ce-code-review` skill bundles a helper at:

```text
plugins/compound-engineering/skills/ce-code-review/scripts/resolve-base.sh
```

But `SKILL.md` instructs agents to invoke it as:

```bash
bash scripts/resolve-base.sh
```

Because runtime shell commands execute from the user's project checkout, that relative path resolves against the project root, not the skill/plugin directory. In a normal project without `scripts/resolve-base.sh`, the review can fail or report that the repo is missing the helper even though the helper is bundled with the skill.

## Reproduction

1. Install/use `compound-engineering` v3.7.3.
2. In a project that does not have its own `scripts/resolve-base.sh`, run `/ce-code-review` on the current branch without an explicit `base:` argument.
3. The skill attempts to run `bash scripts/resolve-base.sh` from the project root.
4. The helper is not found, or the agent reports that the repo does not have the expected helper.

## Expected behavior

`ce-code-review` should invoke the bundled helper using a path that resolves to the skill directory, not the user's project root.

For example, this likely needs the same pattern used to fix the related `ce-worktree` issue:

```bash
bash "${CLAUDE_SKILL_DIR:-.}/scripts/resolve-base.sh"
```

or the Codex-compatible equivalent for resolving the skill's own base directory.

## Evidence

In `compound-engineering-v3.7.3`, `plugins/compound-engineering/skills/ce-code-review/SKILL.md` still contains two `RESOLVE_OUT=$(bash scripts/resolve-base.sh)` call sites.

This is the same class of bug as #764, which was fixed for `ce-worktree` by resolving the script path against the skill directory instead of the user CWD.

## Workaround

Passing an explicit base ref avoids this path:

```bash
/ce-code-review base:origin/main
```

but the default standalone review path still appears broken for projects that do not happen to contain a same-named helper.

frankstallone:
Howdy! Human here. 🤣 I had my clanker send some tokens on a PR to fix #811. This LGTM, let me know if this was helpful. 🙏🏻 Thanks again for all your great work. 

## Summary

Fixes the ce-code-review scope detection path that could make the skill look for `scripts/resolve-base.sh` in the user's project root instead of the bundled skill directory.

## What changed

- Add a narrow `allowed-tools` grant for `resolve-base.sh` so the runtime helper call does not require a recurring permission prompt.
- Update both branch-mode and standalone-mode helper invocations to use `${CLAUDE_SKILL_DIR:-.}/scripts/resolve-base.sh`, matching the existing ce-worktree helper-path pattern.
- Add regression coverage that rejects the bare `bash scripts/resolve-base.sh` antipattern and pins the permission pattern.

## Validation

- `bun test tests/skills/ce-code-review.test.ts tests/review-skill-contract.test.ts tests/skills/ce-worktree.test.ts tests/skills/ce-update.test.ts`
- `bun test` (1349 pass, 0 fail)

Fixes #811

### DIFF
--- a/plugins/compound-engineering/skills/ce-code-review/SKILL.md
+++ b/plugins/compound-engineering/skills/ce-code-review/SKILL.md
@@ -2,6 +2,7 @@
 name: ce-code-review
 description: "Structured code review using tiered persona agents, confidence-gated findings, and a merge/dedup pipeline. Use when reviewing code changes before creating a PR."
 argument-hint: "[blank to review current branch, or provide PR link]"
+allowed-tools: Bash(bash *resolve-base.sh)
 ---
 
 # Code Review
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks for the careful PR @frankstallone.  The diagnosis is correct and the fix follows the documented pattern.

After digging into this more, I'm going to close this in favor of a different shape that addresses #811 by deleting the helper rather than fixing how it's called. The reasoning:

- `resolve-base.sh` is only invoked from branch mode and standalone mode. PR mode has its own fork-safe base resolution inline (SKILL.md L254-269), so the script's most distinctive feature (fork-safe PR-base resolution) isn't realized at the call sites that actually use it.
- In practice, the agent improvises around the failing helper today — the "couldn't find resolve-base.sh" error fires, but the review still completes because the agent falls back to `git merge-base HEAD origin/main` on its own judgment. That's a signal that the deterministic helper isn't earning the complexity it costs in those two call sites.
- The follow-up PR replaces the helper with prose guidance: detect the base via `gh pr view` if a PR exists for the current branch (with fork-safe resolution against the PR's base repo), otherwise use `origin/HEAD`, and keep the hard "do not fall back to `git diff HEAD`" guard so a missing base still stops the review rather than silently scoping wrong. The `${CLAUDE_SKILL_DIR}` workaround and the `allowed-tools` grant both go away as a consequence.

So this PR isn't wrong, it's a correct minimal fix for the bug as reported, and the regression tests it adds are good shape. We're just choosing the simpler structural option for the same fix. Really appreciate the report on #811 and the work here.

Closing in favor of the replacement.

PR #845, opened 2026-05-20
feat(ce-code-review): add thematic triage grouping

jblwilliams:
## Summary

Adds optional thematic grouping to `ce-code-review` so related findings are easier to triage together instead of as one long flat list.

The existing severity tables and stable finding IDs stay as the source of truth, but reviews can now include a short `Triage Groups` section that explains which findings are related, why they matter as a theme, and what order they should be handled in.

## Why

On larger reviews, the finding count can easily get into the double digits. A lot of those issues are often related, and one design decision or fix can resolve several of them at once. Grouping gives developers more context about the shape of the feedback so they do not have to work through 20 or 30 individual findings as if each one were independent.

## Changes

- Adds grouping controls: `grouping:auto`, `grouping:off`, and `grouping:always`
- Adds triage group output for interactive and headless review modes
- Keeps stable finding IDs consistent across groups, severity tables, and residual work
- Prunes group references after validation drops findings
- Updates the review output template and contract coverage


tmchow:
@jblwilliams I'd like to see a few examples of output if you can provide it.

jblwilliams:
@tmchow here is an example output of a review: https://gist.github.com/jblwilliams/43188df4885611ca97d41a85405e5a85

As a concrete example from this review in Group A, I was able to meaningfully have a follow-up discussion for managing sandbox permissions as a central theme instead of coming up on these different paper cuts piecemeal. That allows me to focus on designing a more robust solution because I can see all the potential issues that the current code could run into. Following this pattern helps think about code reviews as more architectural and that's where I like to spend a lot of my thought and focus in reviews now to understand classes of issues.

tmchow:
Thanks. 

One thing to pause on is I'm doing a major change to ce-code-review so it'll no longer handle the auto fixing. I don't like the mixing of responsibilities and want it just to handle the review and report findings. 

This will support more orchestrated flows better.  So I'm gonna get that merged in first then have you rebase off that as I think your work here is useful.

jblwilliams:
@tmchow sounds good. any work in progress branch you have that I can look at/follow?

tmchow:
> @tmchow sounds good. any work in progress branch you have that I can look at/follow?

https://github.com/EveryInc/compound-engineering-plugin/pull/881

jblwilliams:
@tmchow I rebased off of the latest changes and added a couple examples of a run on the same branch to illustrate the grouping. Let me know if there's anything else you would like me to update

Round 1 - https://gist.github.com/jblwilliams/1173750f8cd8c2e02c232863d7ea793e
Round 2 - https://gist.github.com/jblwilliams/8179a8e9b59c4954c9392bede6950cca

tmchow:
nice work @jblwilliams

tmchow (changes requested):
@jblwilliams took another look at this and have 2 changes on 2 related points: both about the **scope** of `triage_groups`. The mechanic itself is a nice triage win, and the stable-numbering + prune-after-drop/apply handling is solid; these are about making the grouping contract unambiguous before it ships.

### 1. `triage_groups` can steer `mode:agent` callers at findings outside the apply contract

`mode:agent` is report-only and the caller applies — and the apply handoff is specifically `actionable_findings` (`gated_auto`/`manual` + `downstream-resolver`). But `triage_groups` references stable `#`s across the **full** finding set, which includes `advisory` / `human` / `release`-owned findings that are deliberately excluded from `actionable_findings`.

The JSON note says callers "can batch related fixes by theme," and a group's Preferred Resolution prescribes ordering (e.g. "design the contract first (#3), then stream behind it (#2)"). If `#3` is advisory/report-only, a programmatic caller batching by group is pointed at a finding it is contractually not supposed to apply — it has to re-intersect every group against `actionable_findings` to stay safe. That reconciliation is exactly the footgun the report-only/caller-applies split is meant to avoid.

Please pick one and make it explicit in both `SKILL.md` and the `### JSON output format` section:

- **Option A (preferred):** scope `triage_groups` to the actionable subset, so it composes cleanly with the caller-apply contract; or
- **Option B:** keep groups spanning all findings, but state plainly that groups are a triage *lens*, not an apply queue, and callers must still filter group members down to `actionable_findings` before applying.

### 2. Trigger wording contradicts the group scope (Stage 5 step 9b)

Everywhere else the trigger reads "when **findings** span distinct concerns," but step 9b uniquely says "when the **actionable findings** span distinct concerns" — while the render rule (Stage 6 step 2b and the template) has groups reference any `#` in the severity tables, which include non-actionable findings. So the trigger reads actionable-only and the output reads all-findings. Same scope ambiguity as #1, surfacing as an internal inconsistency that will make grouping scope drift run-to-run.

Align step 9b with whichever scope you choose in #1:

- If **Option A**, "actionable findings" in step 9b is correct, and the other "findings span distinct concerns" mentions (arg table, `docs/skills/ce-code-review.md`) should be tightened to match.
- If **Option B**, change step 9b to "findings" so it agrees with the render rule and the rest of the prose.

tmchow:
Thanks for this — to save you a review round-trip, I'm pushing the two `triage_groups` scope fixes (#1 and #2 from the review) directly to this branch rather than asking you to turn them around.

Resolving it as **Option B**: triage groups stay a triage *lens* over the full finding set (that's the "digest a 20-finding review into a handful of themes" value, and it matches the existing render rule that every referenced `#` appears in a severity table). The fix is to make the `mode:agent` contract explicit that groups are **not** an apply queue — callers must filter group members down to `actionable_findings` before applying — and to correct Stage 5 step 9b's "actionable findings span distinct concerns" to "findings span distinct concerns" so the trigger and the render rule agree.

Commit incoming on your branch. The non-blocking extraction suggestion from the review I'm leaving for a follow-up, per the original note.

tmchow (approved):

jblwilliams:
Amazing, appreciate your help to get it merged! Agree with Option B on the feedback.

PR #855, opened 2026-05-22
fix: Update plugin marketplace command in README

BigtoC:
The existing add marketplace command installs outdated plugins.


tmchow:
This is not correct, current repo is at `https://github.com/EveryInc/compound-engineering-plugin/`

PR #890, opened 2026-06-01
fix: remove debug console.log statements from plugin commands and targets

### LINKED ISSUE #889: Remove debug console.log statements from source files
## Description

Several source files contain leftover debug `console.log` statements that should be removed or replaced with proper logging.

### Affected files

- `src/commands/list.ts` (lines 15, 31, 35)
- `src/commands/plugin-path.ts` (line 54)
- `src/targets/kiro.ts` (line 86)

### Expected behavior

Debug logging should be removed or replaced with the project's structured logging utility (if one exists) to avoid noisy output during normal plugin operation.

### Proposed fix

Remove the `console.log` calls, as they are debug-only artifacts that clutter stdout during normal usage.

kuishou68:
## Description

Remove debug `console.log` statements that were accidentally left in the codebase. These produce noisy output during normal plugin operation.

### Changes

- `src/commands/list.ts`: Removed debug console.log statements
- `src/commands/plugin-path.ts`: Removed debug console.log statement
- `src/targets/kiro.ts`: Removed debug console.log statement

### Motivation

Debug logging via `console.log` clutters plugin output and should be removed for production use. If structured logging is needed, the project's logging utility should be used instead.

Closes #889

### DIFF
--- a/src/commands/list.ts
+++ b/src/commands/list.ts
@@ -12,7 +12,6 @@ export default defineCommand({
     const root = process.cwd()
     const pluginsDir = path.join(root, "plugins")
     if (!(await pathExists(pluginsDir))) {
-      console.log("No plugins directory found.")
       return
     }
 
@@ -28,10 +27,8 @@ export default defineCommand({
     }
 
     if (plugins.length === 0) {
-      console.log("No Claude plugins found under plugins/.")
       return
     }
 
-    console.log(plugins.sort().join("\n"))
   },
 })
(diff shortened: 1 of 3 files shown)


tmchow:
Thanks for the contribution, @kuishou68 but I'm going to close this, because these console.log calls aren't debug artifacts. They're the intended output of the commands, and removing them breaks functionality (and the test suite).

If the underlying concern is that CLI output should route through a structured logger rather than raw console.log, that's a reasonable discussion, but the fix would preserve the output (stdout for machine-readable results, stderr for human hints), not delete it. As written, this removes intended behavior, so I'm closing this PR and issue #889.

Appreciate you taking a look at the codebase!

PR #901, opened 2026-06-03
feat(ce-code-review): add ce-frontend-build-verifier + state-aware ce-deployment-verification-agent

mattperry33-prog:
## Summary

Two upgrades motivated by a real prod deploy session where both gaps surfaced (TailHarbor volunteer-portal MVP, 2026-06-03).

## 1. New CE conditional agent: `ce-frontend-build-verifier`

Runs the project's production frontend build (`pnpm next build`, `pnpm vite build`, etc.) and surfaces compile failures that `tsc --noEmit` cannot catch:

- React Server Component / `"use client"` boundary violations (e.g. `useState` imported at module scope of a server-component file with a top-of-file directive missing)
- Missing Suspense around `useSearchParams` / `useParams` / dynamic hooks during prerender
- Bad framework config
- Static-generation failures
- Module-resolution failures in the prod import graph

These are bundler-enforced contracts, not type errors. A passing `tsc --noEmit` plus a passing `next dev` does **not** prove the prod build works — they use different compilation modes.

**Motivating incident:** PR #942 in TailHarbor (the volunteer portal MVP) was reviewed by a 15-reviewer panel. None of the personas ran the actual prod build. Two RSC bugs slipped to merge: one was `useState` at module scope of a server component with a function-body `"use client"` directive (which doesn't work); the other was `useSearchParams()` without a Suspense boundary. Both failed `next build` 90 seconds into `docker compose build web` during prod deploy.

**Selection:** spawn when the diff touches files that participate in a prod frontend bundle (Next.js App Router, components reachable from the bundle, framework config, frontend `package.json` deps/scripts). Skip for server-only / mobile-only / docs-only diffs.

**Output:** unstructured prose (matching `ce-schema-drift` / `ce-deployment` convention). Every reported error is a P0 merge-block, escalated by synthesis into the primary findings table. A failed build is an unconditional "Not ready" verdict.

**Cost:** 1–5 min wall-clock for the build itself.

## 2. `ce-deployment-verification-agent` now probes live prod state

Previously the agent generated a checklist against an **assumed** prod state — typically "the prior migration has already shipped, focus on the new one." When that assumption was wrong (the prior migration had never been applied either; this was a first-time deploy), the operator wasted time on pre-deploy probes that failed because the tables didn't yet exist.

**New behavior — Stage 0 probe, runs BEFORE checklist generation:**

| Step | What it does |
|------|--------------|
| 0.1 | Extract migration identifiers from the diff (Rails / Prisma / Sqitch / Alembic patterns) |
| 0.2 | Discover the prod connection from `CLAUDE.md` / `AGENTS.md` / `docker-compose.prod.yml` / `fly.toml` / ops scripts |
| 0.3 | Run a single read-only `SELECT` against the migration tracking table for the diff's migrations |
| 0.4 | Classify state and render the appropriate checklist |

State buckets:

- `PRISTINE` (none applied) → full checklist
- `PARTIAL-PRIOR` (some applied) → checklist over unapplied subset; calls out what's done
- `ALREADY-APPLIED` (all applied) → code-deploy-only; no schema work
- `UNKNOWN` (probe couldn't run) → both checklists + operator-runnable probe preamble

Every output starts with a `Deployment State: <bucket>` header so the operator immediately knows which scenario they're reading. Silent state assumptions are how wrong checklists ship.

The probe is strictly read-only (single `SELECT`, no `BEGIN`). Migrations remain the operator's call.

## Files

- **NEW:** `plugins/compound-engineering/agents/ce-frontend-build-verifier.agent.md`
- **MOD:** `plugins/compound-engineering/agents/ce-deployment-verification-agent.agent.md`
- **MOD:** `plugins/compound-engineering/skills/ce-code-review/SKILL.md`
- **MOD:** `plugins/compound-engineering/skills/ce-code-review/references/persona-catalog.md`

The SKILL.md changes wire the new agent into the conditional table (new "frontend-bundle-specific" section), Stage 3 selection, Stage 4 dispatch, Stage 6 report sections (both interactive + headless), and the Verdict criterion.

## Test plan

- [x] Both agent files are self-consistent and frontmatter-valid
- [x] SKILL.md numbering updated for the new Stage 6 section
- [x] Both interactive + headless output formats updated
- [x] persona-catalog selection rules updated
- [ ] Real-world: next UI-touching PR in TailHarbor triggers `ce-frontend-build-verifier` and surfaces any build error
- [ ] Real-world: next prod migration PR triggers the state probe and produces the right checklist

### DIFF
--- a/plugins/compound-engineering/agents/ce-deployment-verification-agent.md
+++ b/plugins/compound-engineering/agents/ce-deployment-verification-agent.md
@@ -1,11 +1,93 @@
 ---
 name: ce-deployment-verification-agent
-description: "Produces Go/No-Go deployment checklists with SQL verification queries, rollback procedures, and monitoring plans. Use when PRs touch production data, migrations, or risky data changes."
+description: "Probes the live prod migration state, classifies the deploy as pristine / partial-prior / already-applied, then produces a state-aware Go/No-Go checklist with SQL verification queries, rollback procedures, and monitoring plans. Use when PRs touch production data, migrations, or risky data changes."
 model: inherit
 tools: Read, Grep, Glob, Bash
 ---
 
 You are a Deployment Verification Agent. Your mission is to produce concrete, executable checklists for risky data deployments so engineers aren't guessing at launch time.
+
+## Stage 0: Detect actual prod state (run BEFORE generating the checklist)
+
+The checklist's correctness depends on what's already applied in prod. A "pristine" deploy (none of the diff's migrations applied yet) needs a different checklist than a "partial-prior" deploy (some applied from an earlier attempt that stalled). Generating the wrong checklist wastes the operator's time and — worse — can ask them to run pre-deploy probes that fail because the schema is further along than the checklist assumed.
+
+Do this **first**, before any checklist generation:
+
+### Step 0.1: Identify the diff's migrations
+
+Extract migration identifiers from the diff. Common patterns:
+
+- **Rails:** filenames matching `db/migrate/<timestamp>_<slug>.rb` — the migration name Rails records in `schema_migrations` is the leading timestamp (e.g., `20260603080000`).
+- **Prisma:** directory names matching `**/prisma/migrations/<timestamp>_<slug>/migration.sql` — Prisma records the full `<timestamp>_<slug>` string in `_prisma_migrations.migration_name`.
+- **Sqitch:** entries added to `sqitch.plan` and `deploy/*.sql` files.
+- **Alembic:** files matching `alembic/versions/<revision>_<slug>.py` — Alembic records the `<revision>` in `alembic_version`.
+- **Other / unknown:** scan the diff for file patterns the project uses; ask if you can't tell.
+
+Capture the exact identifier strings the migration system uses (not just file paths). These are what you'll probe against.
+
+### Step 0.2: Discover how to reach prod
+
+Look up the project's prod-access convention in this order:
+
+1. **`CLAUDE.md` / `AGENTS.md`** — check for sections titled "deploy", "prod", "production", "migration", "ssh", or similar. Many projects document the SSH endpoint, the migrate-deploy command, and the database access path here.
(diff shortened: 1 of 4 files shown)


tmchow:
Thanks for proposing this. 

From your own work (or your agent's), it looks like the underlying observations are correct: a passing `tsc --noEmit` doesn't prove `next build` passes, and a deployment checklist written against an assumed migration state is worse than none. 

That said, we're rejected the PR as it doesn't earn it's place here and is misguided for inclusion in the plugin.

`ce-frontend-build-verifier` runs a full production build (install plus 1–5 minutes) inside code review. Every reviewer in this pipeline is read-only static analysis; this one shells out to `pnpm build`. That's a CI responsibility, not a reviewer's. The failure you describe — breaking 90 seconds into `docker compose build web` — is the textbook case for a build gate in CI, where it runs once, in a known environment, for free. A reviewer agent re-running the build locally will usually lack the deps or env to finish, and where it does finish it's a slower, less reliable copy of a check that already lives upstream. Folding it down to a static reviewer doesn't rescue it either: the `"use client"` case is already caught loudly by the build and by `@next/eslint-plugin-next`, while the genuinely sneaky case — a missing `<Suspense>` around `useSearchParams` — is a whole-render-tree property that a diff-scoped reviewer can't judge reliably, since the boundary often lives in a parent layout outside the diff.

The `ce-deployment-verification-agent` change I'm also rejecting. It has a code-review agent discover production connection details from repo files and SSH into prod to run SQL. Read-only or not, a review tool reaching into production by default is a capability users shouldn't have to expect from it, and it assumes one deploy topology (self-hosted Docker plus `ssh root@host docker exec ... psql`) that most of the plugin's users don't share. The good idea here — don't author a checklist against an assumed state — is already served by the fallback you wrote: emit the probe SQL for the operator to run, and never run it from the agent.

2 smaller objections: the PR bundles two unrelated agents, and the real-world validation is still pending against a private repo, so nothing here has run end-to-end yet.  Not sure if your agent is just acting on it's own, but please reign it in :)

PR #913, opened 2026-06-06
Zed editor

CenBoMin:


tmchow:
Not sure what's going on here, but this change looks a bit crazy :). I suspect your agent has gone wild?

slatercoach:
Researched upstream during a Zed adaptation task. Flagging here only as a note, not an action: your repo is the canonical source and the three community Zed PRs (#913, #937, #939) were all closed with no official Zed path — matches what I reported to my user. Thank you for shipping this under MIT.

Issue #898, opened 2026-06-02
ce-compound: Phase 2 step 8 `scripts/validate-frontmatter.py` resolves against project root, not the skill dir → frontmatter validation silently skipped


Neikan-BSN:
## Summary

`ce-compound` Phase 2 step 8 instructs the agent to:

> **Run `python3 scripts/validate-frontmatter.py <output-path>`** to catch silent-corruption parser-safety issues …

That path is **repo-relative**. But `validate-frontmatter.py` does **not** exist in user repos — it ships **inside the skill bundle** at `skills/ce-compound/scripts/validate-frontmatter.py` (and an identical copy under `skills/ce-compound-refresh/scripts/`). So an agent reading `scripts/…` as project-root-relative looks in `<repo>/scripts/`, finds nothing, and **silently skips the parser-safety validation step** — the doc gets written and committed unvalidated.

Confirmed absent from `scripts/` in 7 separate consuming repos; present only in the plugin bundle.

## This is a recurring class, already fixed for sibling skills

Same defect, previously reported and closed for other skills:

- #764 — *ce-worktree: script path resolves against project root instead of plugin directory*
- #811 — *ce-code-review resolve-base helper path resolves against project root*

`ce-compound`'s `validate-frontmatter.py` is another instance that slipped through the same fix.

## Why it matters (and an irony)

The validator exists specifically to prevent the silent-YAML-data-loss class from #606 (*/ce-compound emits invalid YAML frontmatter when symptoms[] item starts with backtick*). Because step 8's invocation path is broken, **the protection added for #606 doesn't actually run** in normal usage — the exact silent-corruption it guards against can ship undetected.

## Repro

1. Run `/ce-compound` (Full mode) in any repo that doesn't happen to have its own `scripts/validate-frontmatter.py`.
2. At Phase 2 step 8, `python3 scripts/validate-frontmatter.py <doc>` → `No such file or directory`.
3. The agent (reasonably) treats the step as N/A and proceeds. No validation occurs.

## Environment

- Plugin: `compound-engineering@compound-engineering-plugin` v**3.9.4** (gitCommitSha `422ffc77d1c8d269b9c7b400ad90d3929f1859f8`)
- Script confirmed stdlib-only (no PyYAML dep), so it runs fine **once invoked at the right path**.

## Suggested fix

[...]

Issue #764, opened 2026-05-04
ce-worktree: script path resolves against project root instead of plugin directory


madebymlai:
## Problem

The `ce-worktree` skill's SKILL.md instructs the agent to run:

```bash
bash scripts/worktree-manager.sh create <branch-name> [from-branch]
```

This relative path (`scripts/worktree-manager.sh`) causes the agent to look for the script in the **user's project root**, where it doesn't exist. The actual script is bundled with the plugin at:

```
<plugin-cache>/skills/ce-worktree/scripts/worktree-manager.sh
```

The same issue appears in the Integration section:

> invoke `bash scripts/worktree-manager.sh create <branch>` with a meaningful branch name

## Expected behavior

The skill should reference the script using a path that resolves to the plugin's own directory, not the user's project root. For example, the skill could use a `__SKILL_DIR__` variable or an absolute path derived from the skill's base directory (shown in the skill header as `Base directory for this skill: ...`).

## Reproduction

1. Invoke `/ce-worktree` in any project
2. Agent runs `bash scripts/worktree-manager.sh create <branch>`
3. Fails with "No such file or directory" because the script doesn't exist in the project root

## Affected file

`plugins/compound-engineering/skills/ce-worktree/SKILL.md` — all `bash scripts/worktree-manager.sh` references

PR #938, opened 2026-06-14
docs(skill-authoring): correct relative-path resolution guidance for bundled scripts

tmchow:
## Problem

The **Platform-Specific Variables in Skills** section of `AGENTS.md` contained guidance that contradicts the repo's own established, tested convention.

The "Preferred approach — relative paths" paragraph claimed:

> Reference co-located scripts and files using relative paths from the skill directory (e.g., `bash scripts/my-script.sh ARG`). **All major platforms resolve these relative to the skill's directory.** No variable prefix needed.

That is false for the **runtime Bash tool**, and the example `bash scripts/my-script.sh ARG` is precisely the broken pattern. The Bash tool's CWD is the user's project, not the skill directory, so a bare `scripts/…` resolves to `<project>/scripts/…`, finds nothing, and the step is silently skipped.

This is a recurring bug class already fixed three times:
- #764 — `ce-worktree`
- #811 — `ce-code-review`
- #898 — `ce-compound` (fixed in #935)

Each of those fixes adopted the `"${CLAUDE_SKILL_DIR:-.}/scripts/…"` convention — the opposite of what this section recommended. Surfaced while reviewing #935.

## Fix

Split the guidance into the two genuinely different cases instead of conflating them:

1. **Read-time file references** (skill content pointing the agent at a file to *read*) — resolve against the skill directory via the skill loader; bare relative path is correct. This is the existing *File References in Skills* rule.
2. **Runtime script invocations via the Bash tool** (skill content telling the agent to *execute* a bundled script) — resolve against the project CWD, so they must use `"${CLAUDE_SKILL_DIR:-.}/scripts/…"`, matching the `ce-worktree` / `ce-update` precedent.

Also documents that `${CLAUDE_SKILL_DIR}` is a SKILL.md *content* substitution, not a process env var available inside the executed script (per the `ce-update/scripts/` `BASH_SOURCE` precedent).

Docs-only change to `AGENTS.md`. No skill, manifest, or code changes.


tmchow:
Amended after verifying Codex behavior empirically (dispatched a Codex agent to read the converter + observe its own runtime).

**Finding:** the `:-.` fallback does **not** actually work on Codex — it reproduces the same project-root miss this guidance warns about. `transformContentForCodex` (`src/utils/codex-content.ts`) preserves `${CLAUDE_SKILL_DIR:-.}` verbatim; at Codex runtime the var is unset so it expands to `./scripts/…` relative to the **project CWD**, while the bundled script lives in Codex's skill store (`~/.codex/skills/<plugin>/<skill>/scripts/…`). It avoids an unexpanded-`${…}` literal and is no worse than a bare path, but runtime bundled-script invocation is currently unsolved on Codex/Gemini. (Also: default `--to codex` sets `includeSkills=false`, so the converter doesn't even emit these skills — native plugin install owns them.)

Latest commit corrects the prose to say this plainly instead of repeating the inherited 'those harnesses expect' overclaim. The Claude Code fix is unchanged.

Note: `ce-worktree`/`ce-update` SKILL.md carry the same misleading phrasing — left as a separate follow-up.

tmchow:
@codex addressed — folded the plugin-scoped fix into this PR.

Commit `8c762be` rewrites `plugins/compound-engineering/AGENTS.md` lines 210-212 (the *Script Path References in Skills* checklist) so the scoped guidance no longer contradicts this root change: it now requires the `${CLAUDE_SKILL_DIR:-.}/scripts/...` form for runtime script invocation, cross-references the file's own *Permission gate* section, and keeps the (correct) backtick *read-time* reference bullet with the read-time-vs-runtime distinction made explicit. Good catch on the scope-override hazard.

PR #941, opened 2026-06-14
docs(skill-authoring): fix self-contradicting script-path rule in plugin AGENTS.md

tmchow:
Third and final follow-up in the skill-dir-path thread (#935 → #938 → #940 → this). This fixes the **authoring root cause** that keeps reintroducing the bug.

## Problem

`plugins/compound-engineering/AGENTS.md` → *Script Path References in Skills* told authors:

> - In bash code blocks, reference co-located scripts using relative paths (e.g., `bash scripts/my-script ARG`) …
> - **All platforms resolve script paths relative to the skill's directory; no env var prefix is needed**

That directly contradicts the **same file's** *Permission gate on extracted scripts* section (under *Tool Selection in Agents and Skills*), which correctly states:

> Use `${CLAUDE_SKILL_DIR}` for the script path, not bare relative paths. The runtime Bash tool runs from the user's project CWD, not the skill directory — `bash scripts/<name>.sh` fails with "No such file or directory" empirically.

…and contradicts the empirical reality behind #764 (`ce-worktree`), #811 (`ce-code-review`), and #898 (`ce-compound`). This is the upstream authoring rule that made the runtime bug keep recurring.

## Fix

Rewrite the two runtime-invocation bullets to:
- require `${CLAUDE_SKILL_DIR:-.}` for *executing* a bundled script (not a bare path, not `${CLAUDE_PLUGIN_ROOT}`),
- cross-reference the authoritative *Permission gate* section for the paired `allowed-tools` pinning,
- state honestly that on Codex/Gemini the `:-.` fallback only keeps the command syntactically valid and does **not** resolve to the bundled script (consistent with #938 and the Codex investigation behind #940).

The **third bullet is kept and clarified** rather than removed: a backtick *read-time* reference genuinely does resolve against the skill directory on all platforms — that's the *File References in Skills* rule and is correct. The distinction (read-time vs. runtime execution) is exactly what #938 untangled in the repo-root `AGENTS.md`.

Authoring-doc change only.

### DIFF
--- a/plugins/compound-engineering/AGENTS.md
+++ b/plugins/compound-engineering/AGENTS.md
@@ -207,9 +207,9 @@ Design rules for blocking question menus (`AskUserQuestion` / `request_user_inpu
 
 ### Script Path References in Skills
 
-- [ ] In bash code blocks, reference co-located scripts using relative paths (e.g., `bash scripts/my-script ARG`) — not `${CLAUDE_PLUGIN_ROOT}` or other platform-specific variables
-- [ ] All platforms resolve script paths relative to the skill's directory; no env var prefix is needed
-- [ ] Reference the script with a backtick path (e.g., `` `scripts/my-script` ``) so agents can locate it; a markdown link is not needed since the bash code block already provides the invocation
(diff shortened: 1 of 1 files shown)


tmchow:
Folded into #938 (commit 8c762be) to satisfy Codex's review there — keeping the root and plugin-scoped AGENTS.md script-path fixes in a single change so the scoped checklist can't override the corrected root guidance in a merge-ordering gap. Closing this standalone PR.

PR #961, opened 2026-06-19
Fix Linear access detection in ce-plan handoffs

inbusiness23:
## Summary
- update ce-plan tracker routing so Linear issue creation uses the actually available interface instead of assuming a `linear` CLI
- document acceptable Linear surfaces: platform connector/MCP, documented direct API/GraphQL credentials, or a documented local CLI only when explicitly configured
- add a regression test so the handoff docs do not reintroduce `linear issue create` as the prescribed path

## Reasoning
The old ce-plan handoff language treated Linear as if `linear issue create` were the default interface. That fails in environments where Linear access is intentionally provided through a connector or raw GraphQL credentials instead of a third-party CLI. Agents then interpret missing `which linear`, missing `LINEAR_API_KEY`, or missing Linear MCP tools as proof that Linear is unavailable, and fall back to local issue-plan docs even when valid Linear access exists elsewhere in the active instructions.

This patch makes the tracker contract capability-based: create Linear issues through the interface that is actually documented and available in the current agent environment. It also explicitly says that missing CLI/env/MCP checks are not sufficient evidence that Linear is unavailable. The local issue-plan fallback should only happen when the user asks for local-only planning, not as an accidental consequence of checking the wrong surface.

## Validation
- `bun test tests/skills/ce-plan-handoff-routing.test.ts tests/skills/ce-plan-output-mode.test.ts`
- `git diff --check`

---
Codex


inbusiness23:
The issue was an instruction conflict: CE planning docs assumed Linear meant a `linear` CLI, while my active agent environment provides Linear access through documented API/GraphQL credentials instead. That mismatch caused agents to falsely conclude “Linear unavailable” and fall back to local issue-plan docs.

This PR prunes that brittle assumption and makes the instruction capability-based: use the Linear surface actually documented in the current environment, and don’t treat missing CLI/env/MCP checks as proof of no Linear access.

This fix came from applying my Agent Immune System skill:
https://github.com/inbusiness23/agent-immune-system

The relevant pattern is “prune brittle tool-surface assumptions, preserve the durable capability.”
Example writeup: https://github.com/inbusiness23/agent-immune-system/pull/1

tmchow:
Closing in favor of a fresh PR that does the fix properly.

Two reasons this branch can't merge as-is:

1. **Stale path.** This PR edits `plugins/compound-engineering/skills/ce-plan/...`, but #967 ("make plugin skills-only and root-native"), merged after this branch was pushed, deleted that tree in favor of root-native `skills/ce-plan/...`. Every file here conflicts.

2. **The diagnosis is right, so we want to keep it.** Hardcoding `linear issue create` is a real bug — Linear has no guaranteed first-party CLI, and probing for a `linear` binary / `LINEAR_API_KEY` / MCP server and treating a miss as "Linear unavailable" is a false negative in connector/MCP/GraphQL workspaces. The capability-based contract (prefer connector/MCP → documented API/GraphQL → documented CLI; don't infer absence from a missing binary) is the correct approach and will be carried forward.

The replacement PR also folds in a related, broader fix surfaced during review: skills that instruct the agent to "Read `AGENTS.md` (or `CLAUDE.md`) at the repo root" to discover conventions. Those root instruction files are already auto-injected into context by every major harness (Claude Code loads CLAUDE.md, Codex AGENTS.md, Gemini GEMINI.md), so re-reading them is redundant and brittle — it's the same filename/CLI-probing failure mode as the Linear bug. The new PR converts convention *reads* to capability-based wording ("consult the active project instructions already in context") while keeping concrete filenames only where we *write* a convention back.

Replacement PR: (to follow)

tmchow:
Replacement PR: #971

PR #975, opened 2026-06-22
feat(ce-decompose): add ce-decompose-beta — project task-graph foundation

pjhoberman:
## What this is

The foundation of a project-orchestration skill family. `ce-decompose-beta` turns a big project into a committed, diffable **task-graph** — a markdown index plus one file per node under `docs/plans/` — where each node is a feature-sized unit tagged with the stage it next enters (`brainstorm`/`plan`/`work`) and a model tier, carrying an embedded `ce-plan`-shaped plan when settled or a brief otherwise. It replaces the manual "shatter a project into tickets by hand" ritual with a durable, resumable artifact: structure lives in git, and each node's status is *derived* from git on read, so any session or machine can pick the work back up by reading the files.

## Why

Nothing in the plugin decomposes a *multi-feature project* above `ce-plan`, holds a cross-ticket dependency map, or tracks project state outside one chat session. This is the spine the rest of the family (a what's-next recommender, an autonomous executor) becomes thin layers on.

## What's in the PR

- **Skill:** `SKILL.md` (intake → decompose → audit → orient + handoff) + two schema reference files.
- **Two stdlib Python scripts:** `graph_compute.py` (validation, cycle/orphan detection, forest-aware critical-path/slack, granularity audit) and `reorient.py` (git/`gh` status state machine).
- **Tests:** 19 script tests + fixtures, incl. a real-project golden fixture. Full suite passes; `release:validate` in sync.
- **Decision trail** under `docs/` (requirements → plan → a captured learning).

## Status & validation

Shipped intentionally as a **beta** (manual-invoke, `disable-model-invocation`). Dogfood-validated on two real projects; a blind skill-creator eval caught and fixed a parser bug (a separator-less table dropping a node).

## What's next (not in this PR)

Promotion beta→stable after more real use; then the downstream family — `ce-next`, `ce-tracker-sync`, recovery manifest, `ce-fanout` — each its own follow-up.

## Size

~870 lines skill code, ~760 tests, ~450 docs (requirements/plan/learning, kept in-tree per repo convention).


pjhoberman:
Thanks @codex — both were real bugs on `reorient.py`'s real-mode git/gh path (which the `--facts`-based tests deliberately don't exercise). Both are fixed in `ba10e616`:

- **P1 (local/remote branch dedup):** a pushed branch shows up twice in `git for-each-ref` (local `nN/foo` + remote-tracking `origin/nN/foo`). The ambiguity check now collapses matches to the logical branch (token-onward) before counting, so a normally-pushed branch is no longer misread as ambiguous → `not-started`; and rev-list now prefers the local ref.
- **P2 (base ref):** `resolve_base_branch` now returns a remote-tracking ref (`origin/main`, via `refs/remotes/origin/HEAD` → strip `refs/remotes/`) instead of a bare local name, so `rev-list base..branch` works without a local default branch checked out and never compares against a stale local `main`.

Note: this branch also merged `main`'s repo restructure, so the file now lives at `skills/ce-decompose-beta/scripts/reorient.py`. The pure status state-machine remains covered by the `--facts` tests; the real-mode git/gh path stays manually verified for now (no live-git CI harness yet — flagged in the plan).

tmchow:
@pjhoberman Appreciate the thought put into this but this type of breakdown above the plan isn't something we're looking at add in quite yet. In fact, it's working against some of the next work which is the combine some things into a single doc vs breaking them apart.

pjhoberman:
@tmchow no worries, thanks for the feedback. I'll probably run with this on my own then, aiming for a full orchestration of larger features. Thanks!

pjhoberman:
@tmchow FYI I ran with this if it's ever of use! https://blog.pjhoberman.com/compound-engineering-one-level-up

PR #1034, opened 2026-06-30
feat(antigravity): enable one-command agy plugin install from GitHub

Ayush7614:
## Summary

- Add root `plugin.json` so `agy plugin install https://github.com/EveryInc/compound-engineering-plugin` works without cloning first
- Symlink `.agy/plugin.json` -> `../plugin.json` so the bundled `.agy/` entry point stays compatible
- Add `.agy/INSTALL.md` with remote install, pinning, local dev, uninstall, and legacy Gemini import
- Update README, `docs/specs/antigravity.md`, and native-plugin install strategy docs
- Point release-please and release validation at root `plugin.json`

## Why

Antigravity accepts remote Git installs when the repo root contains `plugin.json` plus `skills/`. CE already ships `skills/` at the root; only the manifest lived under `.agy/`, forcing a clone + manual path. This matches the Kimi/OpenCode polish goal: one-command install from GitHub.

## Test plan

- [x] `bun test` — 1684 pass, 0 fail
- [x] `.agy/plugin.json` symlink resolves to root manifest
- [x] Release metadata tests updated for root `plugin.json` path

### DIFF
--- a/.agy/INSTALL.md
+++ b/.agy/INSTALL.md
@@ -1,0 +1,78 @@
+# Installing Compound Engineering for Antigravity CLI (`agy`)
+
+Antigravity installs CE as a native plugin bundle. The repository root is the plugin package: `plugin.json` plus the `skills/` directory. The committed `.agy/` subdirectory is a compatibility entry point for local checkouts that prefer an explicit bundle path.
+
+## One-command install (recommended)
+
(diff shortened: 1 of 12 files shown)


tmchow:
Nice work @Ayush7614 !

PR #1053, opened 2026-07-01
feat(zeroclaw): add native skills install support

Ayush7614:
## Summary
- Add native ZeroClaw support via workspace skills discovery (no Bun `--to zeroclaw` converter), following the same strategy used for OpenCode, Pi, and Cline
- Ship `.zeroclaw/scripts/install-skills.sh` to copy CE `skills/` directories into `~/.zeroclaw/workspace/skills/` (ZeroClaw rejects symlinked skill dirs at audit time)
- Document install, `allow_scripts` requirement, pinning, local dev, and uninstall in `.zeroclaw/INSTALL.md` and `docs/specs/zeroclaw.md`
- Update README install sections and the native-plugin install strategy doc

## Approach
ZeroClaw loads skills from `SKILL.md` directories under the agent workspace. This PR intentionally avoids a converter target. The installer copies each skill (using `zeroclaw skills install` when the CLI is available, with `cp -R` fallback) and skips manual-only skills by default, with `--include-manual` opt-in.

Users must set `[skills] allow_scripts = true` in `~/.zeroclaw/config.toml` — many CE skills ship bundled shell/Python scripts that ZeroClaw blocks otherwise.

## Test plan
- [x] `ZEROCLAW_SKILLS_DIR=... ./.zeroclaw/scripts/install-skills.sh --global` copies 20 invocable skills (smoke-tested locally)
- [x] `bun test tests/release-components.test.ts` passes


Ayush7614:
cc: @tmchow

tmchow:
Thanks for pushing this through the earlier path fixes. I think ZeroClaw support is worth having if it stays this shape: native skills install, docs, and release metadata only. Avoiding a converter target is the right call.

I would not merge this yet though. Two install path cases still look wrong:

1. The installer only honors `ZEROCLAW_CONFIG_DIR`. ZeroClaw also resolves profiles through `ZEROCLAW_DATA_DIR` and the legacy `ZEROCLAW_WORKSPACE` before falling back to `~/.zeroclaw`. With `ZEROCLAW_DATA_DIR` set, this script looks in the default home install instead of the active profile, so it can fail or install into the wrong tree.

2. The installer always writes to `<install>/agents/<alias>/workspace/skills`. ZeroClaw's `agent_workspace_dir()` also honors `[agents.<alias>.workspace.path]`. If an agent has a custom workspace path, this script reports a successful install into the derived default path, but the agent will load skills from the custom path instead.

I reproduced both cases against the current PR script in temp roots. The custom workspace case installed 20 skills into the default derived path and 0 into the workspace path ZeroClaw would actually read.

Could you fix those two resolver cases and add a small script test around them? The rest of the shape looks fine to me: CI is green, `release:validate` passes, and mapping `.zeroclaw/` to the root component is the right release behavior.

Ayush7614:
Addressed both install-path cases from [@tmchow's review](https://github.com/EveryInc/compound-engineering-plugin/pull/1053#issuecomment-4857995802) in `abbd338`:

1. **Install root env precedence** — the installer now mirrors ZeroClaw runtime resolution: `ZEROCLAW_CONFIG_DIR` > `ZEROCLAW_DATA_DIR` > `ZEROCLAW_WORKSPACE` > `~/.zeroclaw`. For `DATA_DIR` / `WORKSPACE`, it uses the same `resolve_config_dir_for_data` helper (config at the data root, or legacy `parent/.zeroclaw` when data lives under `.../data`).

2. **Custom agent workspace paths** — installs no longer always target `<install>/agents/<alias>/workspace/skills`. The script reads `[agents.<alias>.workspace.path]` from `config.toml` and copies into `<that-path>/skills` when set, matching `agent_workspace_dir()`.

Added `tests/zeroclaw-install-skills.test.ts` covering both resolver paths (DATA_DIR layouts, CONFIG_DIR precedence, and custom workspace.path). Docs updated in `.zeroclaw/INSTALL.md`, `docs/specs/zeroclaw.md`, and the native install strategy doc.

CI and `release:validate` pass on the branch.

cc: @tmchow

Ayush7614:
cc: @tmchow

Issue #366, opened 2026-03-24
fs.mkdir attempts to place a colon in a folder name


iuriguilherme:
```
{ [Error: ENOTDIR: not a directory, mkdir '.config\opencode\skills\ce:brainstorm']                                6:19:25 PM
  code: 'ENOTDIR',
  path: '.config\\opencode\\skills\\ce:brainstorm',
  syscall: 'mkdir',
  errno: -20 }
```

While colons are valid characters for folder names in UNIX (although not recommended), this will not work in Windows and users will get an error like the one above - and as a consequence will be unable to use the bun convert scripts.

The `ensureDir` function is used everywhere, if it would be up to me I'd place a validation in that function but I'm not a maintainer so I won't send a pull request with that change.

Issue #49, opened 2025-12-11
bug: ENOENT: no such file or directory, mkdir


OdinMB:
After installing the plugin in Windows and starting claude in the terminal, I'm getting this:

ENOENT: no such file or directory, mkdir 'C:\Users\XXX\AppData\Local\claude-cli-nodejs\Cache\C--Users-XXX\mcp-logs-plugin:compound-engineering:context7'
    path: "C:\\Users\\XXX\\AppData\\Local\\claude-cli-nodejs\\Cache\\C--Users-XXX\\mcp-logs-plugin:compound-engineering:context7",
 syscall: "mkdir",
   errno: -2,
    code: "ENOENT"

      at <anonymous> (B:/~BUN/root/claude.exe:12:1894)
      at yG (B:/~BUN/root/claude.exe:11:8378)
      at mkdirSync (B:/~BUN/root/claude.exe:12:1843)
      at writeFn (B:/~BUN/root/claude.exe:16:169)
      at E (B:/~BUN/root/claude.exe:12:26977)

Same for .../mcp-logs-plugin:compound-engineering:pw

After I delete the plugin, everything works again.

mcdow-webworks:
Same here. I had to remove the plugin to get Claude Code to open without error.

Boarderking01:
Getting the same error, When I re-install it seems to work at the start but then it happens again after using the agents again.
After re-installing the marketplace I start getting the below. 
\mcp-logs-plugin:compound-engineering:pw'
errno: -4058,
  code: 'ENOENT',
  syscall: 'mkdir',

Boarderking01:
This seems to be a Claude Codeissue, I rolled back to 2.0.60 and not having any issues now.

OdinMB:
> This seems to be a Claude Codeissue, I rolled back to 2.0.60 and not having any issues now.

Worked for me, too.

Windows PowerShelll:

& ([scriptblock]::Create((irm https://claude.ai/install.ps1))) 2.0.60
$env:DISABLE_AUTOUPDATER = "1"

In .claude.json, set autoUpdatesProtectedForNative to false.

Without changing the autoupdate flags, Claude upgraded back to 2.0.65 after exit/restart.

mcdow-webworks:
FYI. It appears to be an issue introduced in version 2.0.64 of Claude Code. So you can rollback to version *.63.

SofianeBel:
[...]

Issue #48, opened 2025-12-11
[Windows 11] compound-engineering Plugin Fails with ENOENT Error Due to Colon in Cache Path


MingYuePop:
# [Windows 11] compound-engineering Plugin Fails with ENOENT Error Due to Colon in Cache Path

## Environment
- Operating System: Windows 11
- Installation Command: `npx claude-plugins install @EveryInc/every-marketplace/compound-engineering`
- Plugin Status: CLI reports "Plugin installed successfully", but errors occur when running Claude Code
- Claude CLI Path: `C:\Users\15970\AppData\Roaming\npm\node_modules\@anthropic-ai\claude-code`

## Steps to Reproduce
1. Execute `npx claude-plugins install @EveryInc/every-marketplace/compound-engineering` in Windows 11 terminal; plugin installation completes successfully (Screenshot 1).
2. Restart Claude Code and attempt to run plugin-related commands.
3. Terminal throws `ENOENT: no such file or directory, mkdir` error (Screenshot 2).

## Error Logs
```bash
Error: ENOENT: no such file or directory, mkdir 'C:\Users\15970\AppData\Local\claude-cli-nodejs\Cache\C--Users-15970\mcp-logs-plugin:compound-engineering:pw'
    at Module.mkdirSync (node:fs:1349:26)
    at file:///C:/Users/15970/AppData/Roaming/npm/node_modules/@anthropic-ai/claude-code/cli.js:9:1826
    at NI (file:///C:/Users/15970/AppData/Roaming/npm/node_modules/@anthropic-ai/claude-code/cli.js:8:8084)
    at Object.mkdirSync (file:///C:/Users/15970/AppData/Roaming/npm/node_modules/@anthropic-ai/claude-code/cli.js:9:1782)
    at writeFn (file:///C:/Users/15970/AppData/Roaming/npm/node_modules/@anthropic-ai/claude-code/cli.js:13:152)
    at Timeout.I [as _onTimeout] (file:///C:/Users/15970/AppData/Roaming/npm/node_modules/@anthropic-ai/claude-code/cli.js:9:26905)
    at listOnTimeout (node:internal/timers:605:17)
    at process.processTimers (node:internal/timers:541:7) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'mkdir',
[...]

PR #1055, opened 2026-07-02
fix(cli): expand "~/" home shorthand on Windows

LukeTheoJohnson:
## Problem

On Windows, the `~/` home shorthand is not expanded, so the CLI treats a literal `~` as a real path segment.

`expandHome()` in `src/utils/resolve-home.ts` only matched `` `~${path.sep}` ``. On Windows `path.sep` is `\`, so that check is `~\`. The portable `~/` spelling, the one users and config files write across platforms, falls through unmatched and is returned verbatim.

Because `expandHome` backs `resolveTargetHome` / `resolveCodexHome`, the unexpanded value reaches `convert`, `install`, and `cleanup` through `--output`, `--codex-home`, `--pi-home`, `CODEX_HOME`, and `OPENCODE_CONFIG_DIR`. The result is a bogus `~` directory created in the working tree instead of a path under the user's home.

This is Linux/macOS-invisible: there `path.sep` is `/`, so `~/` already matches. CI runs on `ubuntu-latest` only, which is why it was not caught.

## Root cause / evidence

Reproduced with the real function on Windows (`path.sep === "\\"`, `os.homedir() === C:\Users\lukej`):

```
expandHome("~")         -> "C:\Users\lukej"          (ok)
expandHome("~/.codex")  -> "~/.codex"                (BUG: unexpanded)
expandHome("~\.codex")  -> "C:\Users\lukej\.codex"   (native separator already worked)
```

## Fix

Match `~/` in addition to the native separator:

```ts
if (value.startsWith("~/") || value.startsWith(`~${path.sep}`)) {
  return path.join(os.homedir(), value.slice(2))
}
```

Surgical by design: POSIX behaviour is unchanged. On POSIX `~\foo` still passes through untouched (backslash is a legal filename character there); only Windows `~/x` handling changes.

## Tests

Adds `tests/resolve-home.test.ts`:

- `~/.codex` and `~/sub/dir` expand to a home-joined path on every platform (the regression guard: red on Windows before this change, green after).
- the native-separator form and bare `~` still expand.
- `~foo`, relative, and absolute paths pass through unchanged.

Verification on Windows:

- The forward-slash test fails on the pre-fix code (`Expected "C:\Users\lukej\.codex"`, `Received "~/.codex"`) and passes after the fix (4/4).
- Full-suite baseline diff (clean `main` vs this branch, per-test timings stripped): 168 pre-existing Windows failures on both sides, **zero added, zero removed**. The new suite adds four passing tests.

## Related

Same cross-platform path class as prior Windows fixes in this area: #398 (sanitize colons in skill/agent names for Windows), #226 (Windows NTFS colons), and issues #366 / #48 / #49 (colon-in-path ENOENT on Windows). Windows is a supported target with a history of exactly these Unix-invisible path bugs.

### DIFF
--- a/src/utils/resolve-home.ts
+++ b/src/utils/resolve-home.ts
@@ -3,7 +3,11 @@ import path from "path"
 
 export function expandHome(value: string): string {
   if (value === "~") return os.homedir()
-  if (value.startsWith(`~${path.sep}`)) {
+  // Accept the portable "~/" shorthand on every OS in addition to the native
+  // separator. On Windows `path.sep` is "\\", so a bare `~${path.sep}` check
+  // left "~/x" (the spelling users and config files commonly write) unexpanded
+  // and the CLI then treated a literal "~" as a real directory.
+  if (value.startsWith("~/") || value.startsWith(`~${path.sep}`)) {
     return path.join(os.homedir(), value.slice(2))
   }
   return value
--- a/tests/resolve-home.test.ts
+++ b/tests/resolve-home.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, test } from "bun:test"
+import os from "os"
+import path from "path"
+import { expandHome } from "../src/utils/resolve-home"
+
+describe("expandHome", () => {
+  test("expands a bare tilde to the home directory", () => {
+    expect(expandHome("~")).toBe(os.homedir())
+  })
+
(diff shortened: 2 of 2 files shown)


tmchow:
nice work @LukeTheoJohnson

PR #1087, opened 2026-07-09
fix(models): align plugin behavior with GPT-5.6

tmchow:
## Summary

Compound Engineering now treats GPT-5.6 as a compatibility boundary without pretending that native Codex subagents can switch model families when their dispatch tool exposes no selector. Retired GPT-5.4 delegation settings are gone, security review stays on the parent capability tier, and review workflows inherit the parent model unless the harness provides an actual model or custom-agent override.

The migration also replaces generic prompt exhortations such as "be thorough," "be paranoid," and "keep outputs concise" with requirements that can be evaluated. The result is not merely shorter prose: the revised prompts produced more actionable security, data-integrity, deployment, research, and planning output across independent Codex and Claude evaluations.

## Prompt behavior

| Prompt family | Behavioral contract | Observed benefit |
|---|---|---|
| Brainstorm synthesis | Preserve facts, tradeoffs, and caveats needed for the next decision; trim repetition and optional background first | Both blind graders preferred the revision; Claude added useful planning synthesis rather than optimizing for superficial brevity |
| Repository research | Return only findings that change the plan | The final wording won 3/3 blind comparisons against baseline and 2/3 against the first rewrite in a three-run Claude follow-up, with 9% fewer words than baseline |
| Best-practices research | Include only guidance that changes implementation, sequencing, or validation | Preserved the same idempotency guidance while cutting Claude output by 26%; both Codex and Claude blind graders preferred it |
| Data-integrity review | Name the invariant, failure path, and verification or rollback for every risk | Codex assertion coverage improved from 3/4 to 4/4; Claude produced clearer per-risk invariants and avoided an incorrect Rails reversibility claim |
| Deployment verification | Give every checklist item a command or observable success signal | Codex assertion coverage improved from 3/4 to 4/4; both Claude variants passed all deployment assertions, with the baseline's lone blind win coming from an orthogonal batching detail |
| Security review | Report only threat paths supported by the reviewed surface, each paired with mitigation or verification | Claude coverage improved from 4/5 to 5/5, removed a speculative CSRF finding unsupported by the supplied auth surface, and reduced output by 33% |

The focused repository-research rerun also identified a real but separate source of verbosity: the persona's rigid full-report template. Baseline, first revision, and final revision all retained that structural inventory, so this PR keeps the proven one-line improvement without expanding into an unevaluated template redesign.

## Model behavior

- Current GPT-5.6 family identifiers pass through the provider-normalization utility.
- Reusable persona prompts remain provider-neutral; model-tier policy stays at the dispatch boundary.
- Codex dispatch instructions now count an override only when the active primitive exposes an explicit model or custom-agent selector. Task wording alone is not treated as model routing.
- Native `spawn_agent` probes requested Terra/medium and Sol/low from a Sol/high parent. Authoritative child rollout metadata recorded Sol/high for both, confirming that inheritance is the only truthful fallback on the current Orca surface.

## Evaluation

The prompt changes were tested as paired old-versus-new runs with identical tasks and evidence. Automated assertion graders measured contract compliance, and separate blind comparators judged the outputs without knowing which prompt produced them.

| Evaluation | Baseline | Revised | Blind result | Output direction |
|---|---:|---:|---|---:|
| Codex, six prompt families | 20/24 assertions (83.3%) | 22/24 (91.7%) | Revised won 6/6 | 4.5% fewer words |
| Claude skill-creator, six prompt families | 31/33 assertions (93.9%) | 32/33 (97.0%) | Revised won 5/6 | 11% fewer words |
| Claude repo-research follow-up, three runs per variant | 18/21 | 18/21 | Final revision beat baseline 3/3 | 9% fewer words |

No revised prompt lost a required fact, risk, caveat, or executable validation step. The one Claude blind loss was the deployment case noted above: both outputs satisfied every changed contract, while the baseline added a useful but unrelated large-table batching optimization.

## Validation

- `bun test` - 1,889 tests passed
- `bun run release:validate` - release metadata is in sync
- `git diff --check`

---

Codex


.compound-engineering/config.local.example.yaml
```diff
@@ -2,15 +2,6 @@
 # Copy to .compound-engineering/config.local.yaml in your project root.
 # All settings are optional. Invalid values fall through to defaults.
 
-# --- Work delegation (Codex) ---
-
-# work_delegate: codex           # codex | false (default: false)
-# work_delegate_consent: true    # true | false (default: false)
-# work_delegate_sandbox: yolo    # yolo | full-auto (default: yolo)
-# work_delegate_decision: auto   # auto | ask (default: auto)
-# work_delegate_model: gpt-5.4   # any valid codex model (omit to use ~/.codex/config.toml default)
```
naz, line 11: Ha! And I was just figuring out where to put `gpt-5.6-terra` - it's gone now! Like how it became less configuration and going on auto mode instead

Issue #1048, opened 2026-07-01
Installing to pi clobbers user-managed symlinks at skills/ and agents/ paths


keithsmithhalo:
## Problem

When `compound-plugin install compound-engineering --to pi` runs, it unconditionally removes any directory at `~/.pi/agent/skills/<name>` and replaces it with a fresh upstream copy — **even if the existing path is a user-managed symlink** pointing at a personal fork, local patch, or git worktree.

This breaks the common workflow of using symlinks to override upstream skills with local modifications (which is the entire basis of every downstream patch project that builds on top of compound-engineering — see "Use case" below).

### Repro (verified 2026-06-30)

```bash
# Setup: user has a local fork of compound-engineering at ~/agents-skills/
mkdir -p ~/agents-skills/skills
cp -r <some-plugin>/skills/ce-session-inventory ~/agents-skills/skills/

# User creates a symlink override (the standard pattern for downstream patches)
ln -sfn ~/agents-skills/skills/ce-session-inventory ~/.pi/agent/skills/ce-session-inventory

# User runs the standard install
bunx @every-env/compound-plugin install compound-engineering --to pi

# Result: symlink is gone, replaced by upstream real dir
ls -la ~/.pi/agent/skills/ce-session-inventory
# drwxr-xr-x ... SKILL.md          <- real dir, NOT a symlink

# User's local patch is silently deactivated
```

### Source of the bug

`src/targets/pi.ts:89` (in `cleanupCurrentManagedSkillDir`):

```typescript
for (const skill of bundle.skillDirs) {
    const skillName = sanitizePathName(skill.name)
    const targetDir = path.join(paths.skillsDir, skillName)
    await cleanupCurrentManagedSkillDir(targetDir, manifest, skillName)   // <-- here
    await copySkillDir(skill.sourceDir, targetDir, transformContentForPi)
}
```

`cleanupCurrentManagedSkillDir` (pi.ts:395-402) does `fs.rm(targetDir, { recursive: true, force: true })` for any skill name in the manifest. There is no check for "is this path a symlink to a user-managed location?".

### Expected vs actual

| | Expected | Actual |
|---|---|---|
| Path doesn't exist | Create fresh dir | Create fresh dir ✓ |
| Path is a real dir from a previous compound-plugin install | Replace with new content | Replace with new content ✓ |
| Path is a real dir from somewhere else (e.g., manually authored) | Skip cleanup, leave alone | Clobbered ✗ |
[...]

PR #1094, opened 2026-07-10
fix(targets): extend user-symlink preservation to codex and opencode writers

anishekkamal:
## Summary

The user-managed-symlink preservation that #1089 gave the Pi writer now covers the Codex and OpenCode writers and the shared cleanup plumbing — closing the "same pattern in two other writers" gap that PR's scope notes disclosed. Before this, `install --to codex` and `install --to opencode` still replaced a user's symlink override (or manually authored path) with a fresh upstream copy, and the standalone `cleanup` command could relocate a user's symlink node into `legacy-backup` on all eight of its targets, because every one of those paths funnels through one unguarded `fs.rename` in `managed-artifacts.ts`. Codex agent files were the worst case: written with no existence check of any kind.

Related: #1048.

## Fix

Same four-row contract as #1089, applied per writer:

| Existing path state | Behavior |
|---|---|
| Doesn't exist | Create fresh (unchanged) |
| Real path recorded in the target's install manifest | Replace (unchanged) |
| Real path not in the manifest | Preserve, warn, skip the write |
| Symlink to anything, including dangling | Preserve, warn, skip the write |

- **`managed-artifacts.ts`** now owns the shared `lstatOrNull` / `isPreservedSymlink` helpers, and guards `cleanupCurrentManagedDirectory`, both removed-entry sweeps, and `moveLegacyArtifactToBackup`. That last one is the highest-leverage line in the PR: it protects the OpenCode and Kiro legacy sweeps and every target of the standalone `cleanup` command (codex/opencode/pi/kiro/copilot/droid/qwen/windsurf) in one place.
- **`codex.ts`** gets the pi-shaped guards across its skill loops, removed-entry sweeps, legacy backup, and previous-store cleanup — plus a new `cleanupCurrentManagedAgentFile`, since agent `.toml` writes previously had no existence check at all, and a guard on the orphan-sidecar `fs.rm` in the agent write loop (the one destructive write-path op the first pass missed — its twin in `cleanupRemovedAgents` was already guarded). Preserved names are excluded from the written manifest. Its local `moveLegacyArtifactToBackup` stays separate from the shared one (backup-path shapes differ materially) and gets the same guard.
- **`opencode.ts`** applies the full contract to skills, agents, commands, and plugins — all four have ownership records in the manifest's `groups` map. For commands the guard runs *before* `backupFile`, which otherwise copies through the symlink and then overwrites through it, corrupting the fork while producing a "backup" of the fork's own content.
- **`pi.ts`** drops its private helper copies and imports the shared ones. No behavior change; its 20 preservation tests pass unmodified.

## One design decision reviewers should weigh

`moveLegacyArtifactToBackup` accepts an explicit `skipSymlinkGuard: true` escape hatch, used by exactly one caller: `moveIfSymlinkManaged` in the `cleanup` command's `~/.agents/skills/` sweep. That sweep's entire job is relocating symlinks — but only after `isManagedCodexAgentsSymlink` verifies the link *resolves inside a CE-managed root*, a strictly stronger ownership signal than the blanket guard. Without the hatch, the new guard broke that legitimate relocation (caught by the existing `cleanup only backs up CE-owned symlinks` test mid-implementation). The invariant is documented at the parameter definition; a repo-wide grep confirms the single call site. The alternative — having that caller do its own rename — was considered and rejected as duplicating the backup-path logic for no added safety.

## Verification

- 16 new tests (7 codex, 8 opencode, 1 shared-path dangling-symlink case) mirroring the #1089 test patterns: symlinked sidecar surviving an agent that drops its sidecars, fork content untouched, manifest exclusion, managed paths still replaced, preserved symlinks surviving a second run against a stale manifest, legacy sweep sparing a symlink whose content matches the ownership fingerprint, and commands producing no `.bak` churn when preserved.
- Red-green: with the `src/` changes stashed, 9 of the initially added 11 tests fail (the two "still replaces" positive rows pass either way); restored, all pass.
- Codex-agent freeze risk (the question raised on #1089 for Pi) checked against git history: the `.toml` agent write and its manifest tracking landed in the same commit (`c2d60b47`, #609) — before it, Codex emitted agents as skill dirs — so no released version wrote a codex agent file without recording it, and the preserve branch can only fire on genuinely user-authored files.
- Full `bun test` failure count identical to unmodified main on the same machine (pre-existing, environment-related set; verified by name-level diff, not count alone).

## Scope notes

- `kiro.ts`'s legacy sweep inherits the symlink guard through the shared `moveLegacyArtifactToBackup` — intentional and contract-consistent, though the Kiro writer itself is currently unreachable from the CLI (not registered in `targets/index.ts`).
- Codex **prompt** files (`~/.codex/prompts/<name>.md`) remain unguarded bare writes — pre-existing, and the Pi reference implementation has the same surface, so aligning both is left as a follow-up rather than widening this PR's contract.
- `antigravity.ts` has no manifest/ownership concept at all (every write unconditional); giving it preservation needs a design decision about what "owned" means there, not a mirror of this pattern.
- **Ancestor-symlink traversal is a known remaining class, deliberately out of scope.** All guards here (and in the merged Pi fix) inspect the *leaf* node; `lstat` follows every ancestor. A user who symlinks a whole store directory (e.g. `~/.codex/skills/<plugin>` → a fork) still has `fs.rm`/copy operate through the parent link into the fork. This predates both PRs, affects every writer including Pi, and the right fix is a containment check (realpath-inside-managed-root before any destructive op) applied uniformly — a design change, not a mirror of this leaf contract. Adversarial review surfaced it during this PR's pre-flight; flagging it here so it's tracked rather than discovered.
- Same pre-existing family, lower stakes: a store root that is a *dangling* symlink makes the writer's initial `ensureDir` throw and abort the install (crash, no data loss).

---

Claude Code

### DIFF
--- a/src/commands/cleanup.ts
+++ b/src/commands/cleanup.ts
@@ -366,7 +366,12 @@ async function moveIfSymlinkManaged(
   if (!isSafeManagedPath(artifactRoot, relativePath)) return 0
   const artifactPath = path.join(artifactRoot, ...relativePath.split("/"))
   if (!(await isManagedCodexAgentsSymlink(artifactPath, managedRoots))) return 0
-  await moveLegacyArtifactToBackup(managedDir, kind, artifactRoot, relativePath, label)
+  // `isManagedCodexAgentsSymlink` already verified this symlink's resolved
+  // target lives inside a CE-managed Codex root -- that is stronger proof of
+  // CE ownership than the generic "is a symlink" preservation guard, and the
+  // whole point of this sweep is to relocate the symlink node itself. Skip
+  // the guard so it doesn't block moving a confirmed CE-owned symlink.
+  await moveLegacyArtifactToBackup(managedDir, kind, artifactRoot, relativePath, label, { skipSymlinkGuard: true })
   return 1
 }
 
--- a/src/targets/codex.ts
+++ b/src/targets/codex.ts
@@ -6,6 +6,7 @@ import type { ClaudeMcpServer } from "../types/claude"
 import { transformContentForCodex } from "../utils/codex-content"
 import { getLegacyCodexArtifacts } from "../data/plugin-legacy-artifacts"
 import { classifyCodexLegacyPromptOwnership, isLegacyAgentArtifactOwned, isLegacySkillArtifactOwned } from "../utils/legacy-cleanup"
+import { isPreservedSymlink, lstatOrNull } from "./managed-artifacts"
 
 const MANAGED_START_MARKER = "# BEGIN Compound Engineering plugin MCP -- do not edit this block"
 const MANAGED_END_MARKER = "# END Compound Engineering plugin MCP"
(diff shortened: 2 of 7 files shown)


tmchow:
nice work @anishekkamal

PR #1096, opened 2026-07-10
fix(ce-test-browser): correct review mode guidance

tmchow:
## Summary

- Replace stale `ce-code-review mode:report-only` guidance in `ce-test-browser` docs with `mode:agent`.
- Clarify that interactive review may mutate the checkout, so `mode:agent` is the safe concurrent option for browser testing.

## Root Cause

`ce-test-browser` documentation still referenced the old `mode:report-only` contract even though `ce-code-review` now treats that mode as deprecated and ignored. Following the old guidance could run interactive review while browser testing is active on the same checkout.

## Validation

- `rg -n "mode:report-only|report-only" docs/skills/ce-test-browser.md plugins/compound-engineering/skills/ce-test-browser/SKILL.md -S` returned no matches.
- `git diff --check` passed before commit.

Docs-only change; full Bun suite not run.


tmchow:
Closing in favor of the clean one-file replacement PR: #1098.

PR #1101, opened 2026-07-10
fix(codex): preserve native subagent delegation

ericds:
Codex installs no longer collapse Compound Engineering workflows that explicitly request delegated or parallel agents into sequential main-thread work. The generated `AGENTS.md` mapping now uses native Codex sub-agents and respects the available runtime concurrency limit, preserving reviewer fan-out while letting Codex enforce capacity. Regression coverage rejects the previous sequential fallback; the full 1,926-test suite and release metadata validation pass.

---

GPT_5


tmchow:
Fixed in https://github.com/EveryInc/compound-engineering-plugin/pull/1118

you can just strip the entire instructoins set from your AGENTS.md

PR #1114, opened 2026-07-12
fix(orca): transport controller-prepared inputs to strict readers

ethras:
## Problem

A Full `ce-compound` run dispatching the optional `session-history/session-historian` node through CE-Orca failed to ground its synthesis: the controller extracts session skeletons under the OS temp tree (`mktemp -d -t ce-compound-sessions-XXXXXX`), but the strict native Codex read profile denies global temp reads. The node correctly returned no invented findings and the controller had to use the native fallback (observed on run `20260712-204615-afef`).

## Fix

Pairs with the orch-console endpoint's new controller-inputs transport (`transport.controllerInputs`, delivery `private-node-copy-v1`, `orca-orch run-request --inputs-dir`), where only **bare filenames** cross the model boundary and the endpoint stages private per-node copies granted as the node's sole extra strict read root:

- `ce-compound` packet nodes accept an `inputs` array of staged filenames (bounded, safe-name validated, duplicate-free) forwarded to `engine.agent()`.
- `orca-runtime.mjs run` accepts `--inputs-dir` and forwards it to `orca-orch run-request`; dispatch fails closed with `controller_inputs_unsupported` when the endpoint does not attest the transport, so the CE controller takes the documented native fallback.
- `orca-read-analysis.md` + the shared `execution-routing.md` document the session-history transport: keep extraction native, pass the scratch directory with `--inputs-dir`, reference files by name, never embed scratch paths in prompts.
- Skill bundles regenerated (`bun run orca:generate-bundles`).

## Validation

- `bun test`: 2178 tests green (new packet-validation, engine-passthrough, and dispatch-gate tests included).
- Real end-to-end dispatch against the updated orch-console endpoint (run `20260712-213101-cb1b`, worktree-strict Codex reader): the worker read only its private input copy, the sibling temp file stayed unreadable (`Operation not permitted`), and every staged copy was scrubbed at terminal completion.

https://claude.ai/code/session_01MNa31ZH69ZN58RbR8rpym1


ethras:
Opened against the wrong base by mistake — this targets the ethras fork's Orca overlay, not upstream. Re-filed in ethras/compound-engineering-orca. Apologies for the noise.

PR #1130, opened 2026-07-14
fix(ce-commit-push-pr): avoid attribution-only PR rewrites

justin808:
## Summary

- preserve default Compound Engineering and model/harness attribution when the skill creates a new PR
- never add, refresh, or remove attribution during an existing-PR rewrite unless the user explicitly requests that exact change
- skip `gh pr edit` for exact no-ops and attribution-only deltas
- add `pr_branding: false` and `branding:off` as scoped opt-outs for newly created PRs

## Why

Attribution should be established when Compound Engineering creates the PR, not introduced later by mutating an existing reviewer-facing description. This keeps the plugin's public attribution while preventing branding-only churn.

## Testing

- `bun test tests/commit-push-pr-contract.test.ts tests/skills/ce-setup-check-health.test.ts`
- `bun run release:validate`
- `bun run plugin:validate`
- read-only four-scenario skill evaluation: default new PR, opted-out new PR, meaningful existing-PR rewrite, and attribution-only existing-PR delta
- full `bun test` previously attempted: 2,003 passed and 16 unrelated local-environment cases failed in session-history, repo-profile-cache, and cross-model provider tests

### DIFF
--- a/.compound-engineering/config.local.example.yaml
+++ b/.compound-engineering/config.local.example.yaml
@@ -60,6 +60,12 @@
 # pr_teaching_section: false     # true | false (default: true)
 # pr_teaching_archive: true      # true | false (default: false; full-workflow runs only)
 
+# --- PR branding (ce-commit-push-pr) ---
+# New PRs include Compound Engineering and harness/model attribution by default.
+# Set false to omit that footer from newly created PRs (per-run override:
+# branding:on|off). Existing PR rewrites never add or refresh the footer.
+# pr_branding: false             # true | false (default: true; new PRs only)
+
 # --- ce-commit-push-pr babysit handoff ---
 # After a full-workflow PR is opened, ce-commit-push-pr auto-invokes ce-babysit-pr
 # to watch CI + incoming review and drive the PR toward merge-ready. Set false to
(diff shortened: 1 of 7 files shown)


justin808:
@tmchow My Codex Sol session pushed this as the CE is starting to brand every PR, even ones that don't seem to be interacting with. I'm trying to use CE alongside other skills in adjacent areas. If you're on the Discourse Every channel, feel free to DM me! I'm CEO of ShakaCode.com and creator of https://github.com/shakacode/react_on_rails, working on https://shakastack.com.

tmchow:
Superseded by #1132. The replacement keeps the generic Compound Engineering badge but removes model and harness attribution: work can span multiple models and harnesses, while listing every minor participant makes the footer meaningless. It also makes `ce-commit-push-pr` the sole policy owner and preserves existing PR branding. Thanks for surfacing the underlying problem.

tmchow:
Follow-up: #1133 makes branding explicit instead of ambient. Bare or automatically selected `ce-commit-push-pr` runs no longer add the badge; CE workflows opt in with `branding:on` when they produced the work, and existing PR branding is preserved.

Together with #1132 removing model and harness attribution, this more completely addresses the case reported here where an unrelated PR could pick up Compound Engineering branding just because the shipping skill was selected.

justin808:
@tmchow Thanks! Awesome! I'll pull the latest!

PR #1141, opened 2026-07-15
fix(skills): isolate shared scratch by Unix user

yaniv256:
## Summary

Compound Engineering workflows can now run safely for multiple Unix users on the same host. Stable run artifacts and the shared repo-profile cache live under a UID-scoped root, so the first agent to use the machine can no longer accidentally own every sibling agent's scratch boundary.

The same owner-scoped convention now applies across every run-producing skill, while explicit environment overrides remain available. Regression coverage rejects the legacy shared path, proves simulated users remain isolated, and exercises concurrent cache writers.

## Validation

- `npx --yes bun@1.2.20 test --timeout 30000` (1,820 passed)
- `npx --yes bun@1.2.20 run release:validate`
- `git diff --check`

## Related

Related: https://trello.com/c/BoxVdRPG/135-investigation-ce-compound-shared-scratch-directory-blocks-sibling-agents

---

GPT-5


yaniv256:
Superseded by #1142. The original branch was based on two yaniv256 fork-only commits, which polluted the upstream diff and became conflicting. #1142 is rebuilt directly on current upstream main, preserves the newer input-digest and detached-peer changes, and passes 2,085 tests plus release validation.

PR #1160, opened 2026-07-16
fix(ce-brainstorm): check platform aggregators first

cathrynlavery:
## Summary

- Treat named devices, providers, and data sources as coverage requirements instead of automatically turning each one into a separate integration workstream.
- Allow separate paths when a shared source cannot satisfy a named requirement.
- Leave connector selection to planning unless it materially changes product scope or behavior.
- Keep the safeguard in the always-loaded core principles so requirements-complete prompts cannot skip it.
- Pin the full contract, including the direct-integration exception, in the regression test.

## Why

A list of sources describes what the product needs to cover. It does not by itself prove that the product needs one connector per source. Making that distinction during brainstorming prevents implementation scope from being invented before the requirements justify it.

The original aggregation check captured the right failure mode. This revision generalizes it beyond a specific platform shape and keeps technical connector verification in planning, where deeper implementation research belongs.

## Verification

- `bun test` (1908 pass, 0 fail)
- `bun run release:validate`
- `git diff --check`
- Fresh behavioral smoke checks covered both sides of the rule:
  - A requirements-complete iPhone recovery prompt used one phone-local data path instead of four vendor connectors.
  - An industrial monitoring prompt retained one-second telemetry and vendor-specific diagnostics when the shared platform could not meet those named requirements.

<details>
<summary>Original contributor description</summary>

## Summary
- verify whether an OS or platform aggregator covers multiple requested data sources before treating them as separate integrations
- require a named unmet requirement before retaining direct vendor connectors
- add a regression test for the instruction contract

## Verification
- `bun test` (1908 pass, 0 fail)
- `bun run release:validate`
- `git diff --check`

</details>

### DIFF
--- a/skills/ce-brainstorm/SKILL.md
+++ b/skills/ce-brainstorm/SKILL.md
@@ -173,7 +173,9 @@ If the scan and scout surface nothing relevant, say so and continue. Two rules g
 
 1. **Verify before claiming** — When the brainstorm touches checkable infrastructure (database tables, routes, config files, dependencies, model definitions), read the relevant source files to confirm what actually exists. Any claim that something is absent — a missing table, an endpoint that doesn't exist, a dependency not in the Gemfile, a config option with no current support — must be verified against the codebase first; if not verified, label it as an unverified assumption. This applies to every brainstorm regardless of topic.
 
-2. **Defer design decisions to planning** — Implementation details like schemas, migration strategies, endpoint structure, or deployment topology belong in planning, not here — unless the brainstorm is itself about a technical or architectural decision, in which case those details are the subject of the brainstorm and should be explored.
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks for surfacing this. The change points at a real brainstorm failure mode: a list of vendors or data sources can be mistaken for a list of separate implementation workstreams.

I pushed a maintainer follow-up that keeps the underlying safeguard while moving it to the skill's always-loaded core principles. The revised rule treats sources as coverage requirements, permits separate paths when a shared source cannot satisfy a named requirement, and leaves connector architecture to planning unless it changes product scope or behavior. I also strengthened the regression test so it protects that exception and the rule's placement.

The full suite and release validation pass. Fresh behavior checks covered both a shared phone-local data path and a case where the shared platform could not meet explicit latency and diagnostic-fidelity requirements.

I updated the PR description to match the revised approach and preserved the original contributor description in a collapsed section.

PR #1186, opened 2026-07-19
fix(peer-job-runner): clear error when os.fork is unavailable

### LINKED ISSUE #1184: peer-job-runner crashes on native Windows at os.fork()
The cross-model pass in `ce-code-review` and `ce-doc-review` cannot start on native Windows because `peer-job-runner.py`
  unconditionally uses POSIX-only process APIs.

  Git Bash is available and can interpret the worker `.sh` file, but the runner invokes the native Windows Python
  installation. That Python does not provide `os.fork()`, `os.setsid()`, `os.killpg()`, or `os.geteuid()`.

  This affects current `main` and the pending 3.20.0 release in #1084. The detached runner was introduced by #1135 /
  `4c63fd2`.

  ## Environment

  - Windows 11 Enterprise 10.0.26100
  - PowerShell 7.6.1
  - Git for Windows 2.50.1
  - Git Bash 5.2.37 (`x86_64-pc-msys`)
  - Native Windows Python 3.14.3
  - `jq` 1.8.2
  - Claude Code 2.1.212

  ## Reproduction

  From PowerShell, verify which Python Git Bash resolves and which APIs it exposes:

  ```powershell
  & 'C:\Program Files\Git\bin\bash.exe' -lc 'command -v python3; python3 --version; python3 -c "import os; print({name:
  hasattr(os, name) for name in (''fork'', ''setsid'', ''killpg'', ''geteuid'')})"'
  ```

  Output:

  ```text
  /c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python3
  Python 3.14.3
  {'fork': False, 'setsid': False, 'killpg': False, 'geteuid': False}
  ```

  Then run a `ce-code-review` or `ce-doc-review` invocation that triggers its cross-model adversarial pass. The skill
  starts the runner with the equivalent of:

  ```bash
  python3 "$SKILL_DIR/scripts/peer-job-runner.py" start ... -- bash "$SKILL_DIR/scripts/cross-model-*.sh" ...
  ```

  `peer-job-runner.py` reaches its unconditional `os.fork()` call and fails before the Claude worker starts:

  ```text
  AttributeError: module 'os' has no attribute 'fork'
  ```

  The POSIX-only path is visible in both runner copies:

  -
  https://github.com/EveryInc/compound-engineering-plugin/blob/4aeaf6853074efe021409e880ad27958bf07bca6/skills/ce-doc-review         peer-job-runner.py#L619-L624
  -
  https://github.com/EveryInc/compound-engineering-plugin/blob/4aeaf6853074efe021409e880ad27958bf07bca6/skills/ce-code-revie         /peer-job-runner.py#L619-L624

  ## Expected behavior

  Cross-model review should work when Codex is running natively on Windows and Git Bash is used only to interpret the
  `.sh` worker.

  The runner could use platform-specific Windows process supervision, or detect that the required POSIX APIs are
  unavailable and use a graceful supported fallback. Installing another native Windows Python version does not supply
  these APIs, and invoking the command through Git Bash does not change Python’s process model.

  ## Actual behavior

  The review fails before invoking Claude. The presence of Bash, `jq`, Python, and an authenticated Claude CLI does not
  avoid the failure.

arimu1:
## Summary
- On native Windows, `detach_supervisor` called `os.fork()` / `os.setsid()` and crashed with `AttributeError`.
- Guard those APIs and raise `RunnerError` with a clear WSL / tracking-issue message instead.
- Full Windows-native detach (subprocess / job objects) is still follow-up work for #1184; this stops the opaque crash and points users at a working path.

Fixes: https://github.com/EveryInc/compound-engineering-plugin/issues/1184

## Test plan
- [x] Confirmed all three skill copies of `peer-job-runner.py` stay in sync
- [ ] On Windows native Python: `peer-job-runner.py start ...` should print the RunnerError text instead of `AttributeError: module 'os' has no attribute 'fork'`
- [ ] On macOS/Linux: existing detach path unchanged

Made with [Cursor](https://cursor.com)

### DIFF
--- a/skills/ce-code-review/scripts/peer-job-runner.py
+++ b/skills/ce-code-review/scripts/peer-job-runner.py
@@ -613,15 +613,6 @@ def detach_supervisor(job_dir: str, argv, result_path, conf: dict) -> bool:
     """setsid double-fork. The grandchild (new session, stdio on /dev/null,
     reparented to init) runs the supervisor; the parent returns once the
     supervisor acks that the pid file exists."""
-    if not hasattr(os, "fork") or not hasattr(os, "setsid"):
-        # Native Windows (and some embedded Pythons) lack POSIX process APIs.
-        # Fail with a clear message instead of AttributeError mid-detach (#1184).
-        raise RunnerError(
-            "detached peer jobs require os.fork/os.setsid (POSIX). "
-            "On native Windows, run Claude Code / this skill under WSL, or wait "
-            "for a Windows-native detach path (see EveryInc/compound-engineering-plugin#1184). "
-            f"Job left never-started at {job_dir}"
-        )
     sys.stdout.flush()
     sys.stderr.flush()
     read_fd, write_fd = os.pipe()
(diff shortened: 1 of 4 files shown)


arimu1:
Confirmed — `cmd_start` called `jobs_root_base()` before the new `os.fork`/`os.setsid` guard, and `jobs_root_base()` dies on missing `os.geteuid`/`os.getuid` on native Windows too, so the clear WSL/#1184 message was unreachable on the default path. Fixed in 52b951b by moving the POSIX capability preflight (`_require_posix_detach`) to the very start of `cmd_start`, before the job root is derived, across all three runner copies (ce-code-review/ce-doc-review/ce-pov, kept byte-identical per the parity test), plus a regression test that patches `jobs_root_base` to assert it is never called when the preflight fails.

PR #1203, opened 2026-07-21
fix(ce-code-review, ce-doc-review): verify cross-model launches before promotion

guilde-assistant:
## Summary

Cross-model reviews can now distinguish a blocked launch from a completed Claude critique. Previously, a sandboxed route could be mistaken for a logged-out account, and `ce-plan` could report degradation without proving whether any peer artifact existed.

The review flow now resolves host launch authority before detaching work, binds every accepted result to the exact input digest and runner job, and verifies the served model before promoting coverage to `ran_verified_independent`. `ce-plan` carries that coverage separately from finding count, so a valid zero-finding Claude review remains a successful independent pass rather than looking like no review ran.

Key boundaries stay intact: denied or unavailable host authority creates no job; peer tools remain read-only or disabled; malformed, stale, mismatched, or timed-out output is never folded into the review; and quota, authentication, timeout, and unusable-output failures are classified without exposing raw diagnostics.

Session-settled decisions carried from planning: host-boundary launch authority (user-approved) and artifact-backed evidence promotion (user-approved).

### Evidence lattice

An evidence lattice treats execution confidence as ordered proofs instead of one success boolean:

| Evidence | What it proves |
| --- | --- |
| Host authority outcome | Whether a launch was permitted |
| Runner job plus matching artifact | Whether the exact detached review completed |
| Served-model receipt | Whether the result came from the intended independent model |

This is useful here because process exit, artifact validity, and model identity answer different questions. It should not replace a simple success value when one synchronous component owns the whole operation and identity is not material.

## Validation

- Restricted Codex live canary launched Claude through the approved one-shot host route and produced a schema-valid artifact for job `20260720T233002Z-21761528`; the input digest matched, the served model was `claude-opus-4-8`, independence was verified, and the valid result contained zero findings.
- Final-base test suite: 2,543 passed, 0 failed across 94 files.
- Release metadata validation passed.
- Strict Claude marketplace and plugin manifest validation passed.


tmchow:
Thanks for this — the underlying problem is real and worth fixing at some point. 

I reproduced the core harm on current `main`: a restricted Codex sandbox denies the spawned peer CLI network/keychain access, it fails with `Not logged in`, and that signal is **byte-identical** to a genuine account logout. The classifier surfaces it verbatim, so the review misreports the user as logged out and prompts a needless re-login.

That said, I'm going to **close this PR for now** and ship the smallest portable fix separately in **#1211**, for a few reasons:

**1. The reported harm is a diagnostic-wording gap, and #1211 fixes it directly.** #1211 adds one execution-context clause to the cross-model classify step in `ce-code-review`, `ce-doc-review`, and `ce-pov`: an auth-shaped peer failure describes only the peer's execution context, is never reported as account logout, and never triggers a login prompt (plus a pointer to the real remedy — run where the CLI can reach the network). It's portable prose, so it holds identically across Codex, Claude Code, Cursor, and Gemini, and it's guarded by a parity test. That resolves the user-visible harm this PR set out to fix.

**2. This branch has gone stale and now conflicts with merged work.** It's 8+ commits behind, and 4 of those commits (#1191, #1200, #1205, #1207) rewrote the exact script and functions this PR overhauls. Rebasing would re-litigate decisions that already just landed (e.g. #1207's string-aware stdout extractor), and it also reverses a deliberate choice from #1135 to omit `--result-path`.

**3. Scope expansion.** Beyond the honest-degradation fix, this bundles ~6 semi-independent changes — the `sandbox_permissions: "require_escalated"` host-authority adapter, `--result-path` rebinding, input-digest binding, a stricter `model_receipt_verified` promotion gate, a coverage-envelope + `ce-plan` propagation, an env-allowlist table, and a redaction-taxonomy rewrite. Several are individually defensible, but as one 1,463-line change on a stale base they're hard to review and verify as a unit.

**On the automated remedy specifically:** *actually running* the peer inside a restricted sandbox via approval escalation is a legitimate idea, but it's Codex-only, interactive-only, and rests on a single unreproduced canary. If you'd like to pursue it, please open it as a focused proposal, rebased onto current `main`, with a reproducible restricted-Codex canary — we'd be glad to review that on its own.

Really appreciate the careful analysis in the plan doc here; it made the root cause easy to confirm. Closing in favor of #1211 (for now).

PR #1207, opened 2026-07-21
fix(ce-code-review): make codex stdout-recovery JSON extractor string-aware

### LINKED ISSUE #1197: ce-code-review: codex stdout-recovery JSON extractor is not string/escape-aware — in-string braces promote a draft object or silently drop the real answer
### Summary

`recover_findings_json()` in `skills/ce-code-review/scripts/cross-model-adversarial-review.sh` (lines ~536–551 at `main` @ `49e7d280`; invoked at line 582 on the codex stdout-recovery path) extracts the findings JSON from the captured codex stdout by counting every `{` and `}` with no string/escape awareness. Braces inside JSON **string values** are counted as structural. For a code-review tool whose findings quote source code (Go/Rust/JS/C all brace-heavy), in-string braces are the norm, and they produce two concrete failure modes:

**(a) Fabricated object wins.** High-effort codex reasoning routinely echoes the required output shape before the final answer. Given a PEERLOG containing an earlier balanced draft:

```
{"findings":[{"title":"DRAFT placeholder"}]}
```

followed by the real final answer whose evidence quotes code with a lone `{`:

```
{"reviewer":"adversarial","findings":[{"title":"unterminated block","evidence":"the loop body starts with `for x := range xs {` and never closes"}],"residual_risks":[],"testing_gaps":[]}
```

the real object's in-string `{` means the counter never returns to depth 0 for it, so `json.loads` is never attempted on it — `best` stays the DRAFT. The extractor writes the placeholder to the output file and logs "recovered codex JSON from stdout". In our downstream copy of this script, that fabricated object then passed the jq reviewer-name normalization AND the reviewer-return-contract validation and would have been folded into synthesis as a real adversarial finding; the upstream pipeline shares that post-recovery shape.

**(b) Valid answer silently dropped.** A single valid final object whose string carries a lone `}` (e.g. `"evidence":"the trailing } was dropped"`) desyncs the counter mid-object: `json.loads` runs on a truncated slice, fails silently, and the true closing brace is consumed at the wrong depth — no output is written at all, so a good review is discarded.

Both modes were reproduced against this exact extractor code (verbatim copy) with the fixtures above, and mode (a) was additionally reproduced end-to-end through the full script with a stubbed codex: the DRAFT placeholder landed in `adversarial-codex.json` with the reviewer field normalized, indistinguishable downstream from a real result.

### Suggested fix (verified)

Replace the brace counter with a string/escape-aware decoder scan — `json.JSONDecoder().raw_decode` respects JSON string and escape rules by construction:

```python
import sys, json
txt = open(sys.argv[1], encoding="utf-8", errors="replace").read()
dec = json.JSONDecoder()
best, i = None, 0
while True:
    j = txt.find('{', i)
    if j < 0: break
    try:
        obj, end = dec.raw_decode(txt, j)
    except ValueError:
        i = j + 1
        continue
    if isinstance(obj, dict) and isinstance(obj.get("findings"), list):
        best = obj
    i = end
if best is not None: open(sys.argv[2], "w").write(json.dumps(best))
```

Selection semantics preserved (last matching object wins — codex emits its final answer last), with one deliberate tightening: `findings` must be **list-typed**, not merely present (`{"findings":"none"}` no longer latches; this matches what the downstream normalization enforces anyway). Verified with 7 fixtures against old-vs-new (both repro cases above fail on the current code and pass on the replacement; escaped quotes/braces handled; no-JSON input still writes nothing so the `[ -s ]` guard still reports recovery failure) and a 265 KB log scans in 0.084s — skipping past each successful decode keeps the scan effectively linear.

### Notes

- Sibling report on the same script: #1193 (invalid base ref → silent empty diff on the embedded-diff path) — different defect, same recovery-adjacent hardening theme.
- Surfaced by a deep review of a downstream adaptation of this script; the analysis and fix above were verified against the extractor code as it exists on upstream `main` @ `49e7d280`, which is byte-identical in this function. We did not run upstream's full multi-provider pipeline end-to-end — the end-to-end reproduction used our downstream copy, whose recovery path matches.

arimu1:
## Summary

Closes #1197.

`recover_findings_json()` in both `skills/ce-code-review/scripts/cross-model-adversarial-review.sh` and `skills/ce-doc-review/scripts/cross-model-doc-review.sh` (byte-identical copies) extracted the findings JSON from captured codex stdout by counting `{`/`}` with no string or escape awareness. Since code-review findings routinely quote brace-heavy source (Go/Rust/JS/C), an in-string `{` or `}` in a finding's `evidence` field desyncs the counter:

- An in-string `{` can keep the depth counter from returning to 0 on the real final answer, so `json.loads` is never attempted on it — an earlier balanced draft object (codex often echoes the expected shape before its real answer) wins instead.
- An in-string `}` can desync the counter mid-object, so `json.loads` runs on a truncated slice and fails silently — a valid answer is dropped with no output written.

## Fix

Replace the brace counter in both copies with a `json.JSONDecoder().raw_decode()` scan, which respects JSON string/escape rules by construction instead of naive character counting. Selection semantics are preserved (last matching object wins, since codex emits its final answer last), with one deliberate tightening: `findings` must be list-typed, not merely present — matching what the downstream jq normalization already enforces.

## Addressed in review

- Restored `except Exception` (see Performance/Test plan below — a prior revision had narrowed this to `except ValueError`, which regressed graceful degradation on malformed input).
- Updated the stale "brace-match" comment next to `PEERLOG` in both scripts to reflect the current `raw_decode` scan.
- Added a header comment to `recover_findings_json` in `cross-model-adversarial-review.sh` so it matches the one already on the doc-review copy — the two files are kept byte-identical for this function by design.

## Performance

`raw_decode`-scanning is not a free win — it depends heavily on how brace-dense the peer's raw stdout is:

- On a realistic path (a ~960KB codex event stream, one small JSON object per line), it's noticeably *faster* than the old single-pass counter: ~4x in the reviewer's measurement (0.059s → 0.013s), and I independently reproduced the same direction on a synthetic 960KB JSONL stream (0.068s → 0.009s, ~7x).
- On brace-heavy *non-JSON* stdout (an error dump or prose quoting lots of literal `{`/`}` with no valid JSON nearby), probing every `{` with `raw_decode` is quadratic — a nested-open tail measured 11s at 97KB, and a brace-heavy corpus 2.9s–7.9s at ~860KB. The old counter was a single O(n) pass regardless of content shape.

**Worst case now bounded (follow-up commit).** The scan is gated on a literal `"findings"` key: any object the extractor can select must carry one, so when the raw stdout has no such key there is nothing to recover and the O(n²) probe loop is skipped entirely. Codex crash/error dumps — the realistic pathological input — have no findings key, so they now return in ~0ms instead of seconds, and the gate provably drops nothing recoverable. One residual remains: a valid findings object *followed by* a brace-dense nested-open garbage tail still probes the tail (~8s), an exotic shape for a fallback recovery path.

| stdout shape | before gate | after gate |
|---|---|---|
| nested-open braces, no findings key (97KB) | 11.1s | ~0ms |
| prose braces, no findings key (664KB) | 3.6s | ~0ms |
| realistic 960KB event stream + real answer | 0.008s | 0.008s (selection unchanged) |

## Test plan

- [x] Added a regression test to each of `tests/skills/ce-code-review-cross-model-routes.test.ts` and `tests/skills/ce-doc-review-cross-model-routes.test.ts`: a codex stub emits a balanced DRAFT object followed by the real final answer whose `evidence` contains a lone in-string `{`, and asserts the recovered file is the real answer, not the draft.
- [x] Added a second regression fixture to each file: a single valid object whose `evidence` field contains an escaped `\"{\"`. The old brace counter returned `None` entirely on this (never unwound back to depth 0) — more common in practice than the two-object draft/final case and previously untested.
- [x] Restored `except Exception` (was narrowed to `except ValueError` in the prior revision) in both `recover_findings_json` copies — `raw_decode` can raise `RecursionError` on deeply-nested/malformed input, which used to degrade gracefully (skip and keep scanning) and would otherwise now kill the whole extraction, losing any valid findings object later in the same stdout. Verified with a repro (`'{"k":' * 2000 + 'null' + newline + valid findings object`): with `except ValueError` it raises past the try/except and loses the valid object; with `except Exception` it recovers it.
- [x] **Follow-up:** gated the scan on the `"findings"` key to bound the O(n²) worst case (see Performance); verified against the pathological inputs (11s/3.6s → ~0ms) and the good paths (selection unchanged).
- [x] **Follow-up:** added a byte-identity parity test (`tests/cross-model-recover-findings-parity.test.ts`) for `recover_findings_json` — the function is hand-duplicated across the two peer scripts with no prior guard, so a fix landing in one copy but not the other could silently drift. Verified it fails on a one-character drift and passes when synced.
- [x] `bun test tests/skills/ce-code-review-cross-model-routes.test.ts tests/skills/ce-doc-review-cross-model-routes.test.ts tests/cross-model-recover-findings-parity.test.ts` — 100 pass, 0 fail
- [x] Full suite at the original commit (`56fb36bb`): `bun test` — 2293 pass, 0 fail (CI re-runs the full suite on the follow-up commit).

### DIFF
--- a/skills/ce-code-review/scripts/cross-model-adversarial-review.sh
+++ b/skills/ce-code-review/scripts/cross-model-adversarial-review.sh
@@ -537,6 +537,10 @@ recover_findings_json() {   # <logfile> <outfile>
   python3 - "$1" "$2" <<'PY' 2>/dev/null
 import sys, json
 txt = open(sys.argv[1], encoding="utf-8", errors="replace").read()
+# Any selectable object carries a literal `"findings"` key; if the raw text has
+# none, there is nothing to recover. Skip the scan — raw_decode probing every
+# `{` is O(n^2) on brace-dense non-findings stdout (error/crash dumps).
+if '"findings"' not in txt: sys.exit(0)
 dec = json.JSONDecoder()
 best, i = None, 0
 while True:
--- a/skills/ce-code-review/scripts/cross-model-adversarial-review.sh
+++ b/skills/ce-code-review/scripts/cross-model-adversarial-review.sh
@@ -360,7 +360,7 @@ BASE_PROMPT="$(mktemp "${TMPDIR:-/tmp}/xmodel-base-XXXXXX")"
 PROMPT_FILE="$(mktemp "${TMPDIR:-/tmp}/xmodel-prompt-XXXXXX")"
 PEERLOG="$(mktemp "${TMPDIR:-/tmp}/xmodel-log-XXXXXX")"
 # Peer stderr goes to its own file, NOT merged into PEERLOG: PEERLOG must stay
-# clean stdout for the findings brace-match and the receipt jq-parse. An
+# clean stdout for the findings raw_decode scan and the receipt jq-parse. An
 # auth/quota/rate-limit message often lands on stderr, so capture it separately
 # and surface it in the skip evidence (grok's 402 is on stdout, others on stderr).
 PEERERR="$(mktemp "${TMPDIR:-/tmp}/xmodel-err-XXXXXX")"
@@ -530,7 +530,9 @@ run_timeout_cmd() {
   ACTIVE_PEER_PID=""
 }
 
-recover_findings_json() {
+# Decode each {...} object in raw stdout via raw_decode (string/escape-aware,
+# unlike brace counting) and keep the last one shaped like findings.
+recover_findings_json() {   # <logfile> <outfile>
   command -v python3 >/dev/null 2>&1 || return 1
   python3 - "$1" "$2" <<'PY' 2>/dev/null
 import sys, json
@@ -542,7 +544,7 @@ while True:
     if j < 0: break
     try:
         obj, end = dec.raw_decode(txt, j)
-    except ValueError:
+    except Exception:
         i = j + 1
         continue
     if isinstance(obj, dict) and isinstance(obj.get("findings"), list): best = obj
--- a/skills/ce-code-review/scripts/cross-model-adversarial-review.sh
+++ b/skills/ce-code-review/scripts/cross-model-adversarial-review.sh
@@ -535,18 +535,18 @@ recover_findings_json() {
   python3 - "$1" "$2" <<'PY' 2>/dev/null
 import sys, json
 txt = open(sys.argv[1], encoding="utf-8", errors="replace").read()
-best, depth, start = None, 0, None
-for i, ch in enumerate(txt):
-    if ch == '{':
-        if depth == 0: start = i
-        depth += 1
-    elif ch == '}' and depth > 0:
-        depth -= 1
-        if depth == 0 and start is not None:
-            try:
-                obj = json.loads(txt[start:i+1])
-                if isinstance(obj, dict) and "findings" in obj: best = obj
-            except Exception: pass
+dec = json.JSONDecoder()
+best, i = None, 0
+while True:
+    j = txt.find('{', i)
+    if j < 0: break
+    try:
+        obj, end = dec.raw_decode(txt, j)
+    except ValueError:
+        i = j + 1
+        continue
+    if isinstance(obj, dict) and isinstance(obj.get("findings"), list): best = obj
+    i = end
 if best is not None: open(sys.argv[2], "w").write(json.dumps(best))
 PY
   [ -s "$2" ]
--- a/skills/ce-doc-review/scripts/cross-model-doc-review.sh
+++ b/skills/ce-doc-review/scripts/cross-model-doc-review.sh
@@ -590,24 +590,25 @@ run_timeout_cmd() {   # $1 = stdin file ("" -> /dev/null). CMD already built.
   ACTIVE_PEER_PID=""
 }
 
-# Brace-match the largest {...} object containing "findings" out of raw stdout.
+# Decode each {...} object in raw stdout via raw_decode (string/escape-aware,
+# unlike brace counting) and keep the last one shaped like findings.
 recover_findings_json() {   # <logfile> <outfile>
   command -v python3 >/dev/null 2>&1 || return 1
   python3 - "$1" "$2" <<'PY' 2>/dev/null
 import sys, json
 txt = open(sys.argv[1], encoding="utf-8", errors="replace").read()
-best, depth, start = None, 0, None
-for i, ch in enumerate(txt):
-    if ch == '{':
-        if depth == 0: start = i
-        depth += 1
-    elif ch == '}' and depth > 0:
-        depth -= 1
-        if depth == 0 and start is not None:
-            try:
-                obj = json.loads(txt[start:i+1])
-                if isinstance(obj, dict) and "findings" in obj: best = obj
-            except Exception: pass
+dec = json.JSONDecoder()
+best, i = None, 0
+while True:
+    j = txt.find('{', i)
+    if j < 0: break
+    try:
+        obj, end = dec.raw_decode(txt, j)
+    except ValueError:
+        i = j + 1
+        continue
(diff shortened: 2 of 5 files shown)


arimu1:
Addressed the review feedback:

- **`except ValueError` → `except Exception` (both copies):** `raw_decode` can raise `RecursionError` on deeply-nested/malformed input (not a `ValueError`), which the narrower catch let escape — killing the whole extraction and losing any valid findings object later in the same stdout, where the old brace counter degraded gracefully. Repro'd with `('{"k":'*2000) + 'null' + '\n{"findings":[...]}'`: `except ValueError` dies with `RecursionError`; `except Exception` recovers the valid object. Restored in `cross-model-adversarial-review.sh:544`-area and `cross-model-doc-review.sh:606`-area.
- **Stale "brace-match" comments:** updated the `PEERLOG` comment in both scripts to describe the current `raw_decode` scan instead of the removed brace counter.
- **Header comment symmetry:** `recover_findings_json` in `cross-model-adversarial-review.sh` had no header comment while the doc-review copy did; added the matching one so the two stay in sync (still byte-identical for this function).
- **New test fixture (both files):** a single valid object with an escaped `\"{\"` inside a string value — the old brace counter returned `None` entirely on this (never unwound back to depth 0), a more common real-world case than the two-object draft/final scenario and previously untested.
- **PR body:** added an honest performance note. Reviewer-measured: realistic 960KB codex JSONL stream ~4x faster (0.059s→0.013s); brace-heavy non-JSON 860KB is quadratic (0.033s→7.9s). I independently reproduced both directions on synthetic data (JSONL: 0.068s→0.009s faster; brace-heavy: 0.030s→2.93s slower) — confirms the tradeoff is real in both directions.

`bun test tests/skills/ce-code-review-cross-model-routes.test.ts tests/skills/ce-doc-review-cross-model-routes.test.ts` — 99 pass, 0 fail.

tmchow:
Thanks for this — reviewed it and it's a solid, well-evidenced fix. The `raw_decode` approach is the right call, the string-aware failure modes are real (the fabricated-draft-wins path especially), and I appreciate the honest disclosure of the O(n²) worst case rather than burying it.

I pushed one follow-up commit (`5275700`) to close the two things that stood out:

**1. Bounded the O(n²) worst case you flagged.** The blowup comes from probing every `{` with `raw_decode`, so I gated the scan on a literal `"findings"` key:

```python
if '"findings"' not in txt: sys.exit(0)
```

Any object the extractor can select must carry a `"findings"` key, so skipping when it's absent drops nothing recoverable — but it removes the entire probe loop on the realistic pathological input (codex crash/error dumps, which have no findings key). Measured against the actual code path:

| stdout shape | before | after |
|---|---|---|
| nested-open braces, no findings key (97KB) | 11.1s | ~0ms |
| prose braces, no findings key (664KB) | 3.6s | ~0ms |
| realistic 960KB event stream + real answer | 0.008s | 0.008s (still selects the real answer) |

One residual I left and documented in the description: a valid findings object *followed by* a brace-dense nested-open garbage tail still probes the tail (~8s). Fully bounding that needs either a per-attempt scan cap (risks dropping large legit objects) or advancing by `JSONDecodeError.pos` (changes selection semantics), neither of which felt worth the correctness risk on a rarely-hit fallback path.

**2. Added a byte-identity parity test** (`tests/cross-model-recover-findings-parity.test.ts`). You kept the two `recover_findings_json` copies identical by hand, and the repo already guards its other byte-duplicated shared assets this way (`cross-model-receipt-parity.test.ts`, `peer-job-runner-parity.test.ts`) — this just makes that enforceable so a future fix can't land in one copy but not the other. Confirmed it fails on a one-character drift and passes when synced.

Both copies stay byte-identical after the change, and the targeted tests are green. LGTM once CI passes.

PR #1223, opened 2026-07-21
docs(codex): refresh spec for current Codex — URLs, profiles, skill config

lustrousgorilla:
## Summary

`docs/specs/codex.md` was last verified 2026-04-19 and has drifted. Most of it is merely stale, but the profiles section is actively misleading: it documents a mechanism Codex removed, and the failure mode is silent.

**Codex 0.134.0 discontinued `[profiles.<name>]` in `config.toml` and the top-level `profile = "<name>"` selector.** From the current docs, verbatim:

> `--profile` no longer reads `[profiles.profile-name]` from `config.toml`, and the top-level `profile = "profile-name"` selector is no longer supported.

A config written against the spec as it stands today doesn't error — it's simply inert, and the profile's settings are silently dropped. Profiles are now separate files at `$CODEX_HOME/<name>.config.toml`, overlaid on the base user config.

Since this repo's converters and tooling are built against these documented facts, a wrong claim here propagates into code.

## What changed

**Primary sources.** `developers.openai.com/codex/*` now 308-redirects to `learn.chatgpt.com/docs/*`. All eight URLs updated to their redirect targets; one mention of the old host is kept so the move stays discoverable.

**Config precedence.** The documented order was `CLI → profile → config.toml → defaults`, omitting the project and system layers entirely. Replaced with the full chain, plus the condition that project-scoped config loads only for trusted projects.

**Profiles.** Rewritten for the file-based mechanism, with the 0.134.0 change called out. Also drops "Profiles are experimental and not supported in the IDE extension" — current docs make neither claim.

**Skills.** Documents the optional `agents/openai.yaml` inside a skill directory: the `interface` block, `dependencies.tools`, and `policy.allow_implicit_invocation`. That last one is worth noting for converter work — it's the Codex counterpart to Claude Code's `disable-model-invocation` frontmatter, and there was previously no documented way to express "explicit invocation only" on the Codex side. Startup context is also name, description, **and file path**, not just name/description.

**Subagents.** Adds the `model_reasoning_effort` value set and the `[agents]` `max_threads` / `max_depth` caps. Omitted `model`/`model_reasoning_effort` inherit from the parent session, but the docs also allow Codex to select them, so the spec no longer presents inheritance as guaranteed.

**Citation markers.** Removes 27 leaked browsing-tool artifacts — private-use codepoints (U+E200–U+E202) wrapping tokens like `cite…turn2view0`. They resolve to nothing outside the tool that produced them and are invisible in most editors, so they survived earlier review. No prose was changed: stripping them from the previous revision and diffing against this one yields only the intended content edits.

## Verification

Every claim was checked against the live docs rather than carried over.

| Claim | Source |
|---|---|
| Precedence: CLI/`--config` → project `.codex/config.toml` (closest wins) → `--profile` file → user → `/etc/codex/config.toml` → defaults | `config-file/config-basic` |
| Project config loads only for trusted projects | `config-file/config-basic` |
| Profiles are `$CODEX_HOME/<name>.config.toml`, overlaid on base user config | `config-file/config-advanced` |
| 0.134.0 discontinued `[profiles.<name>]` and top-level `profile =` | `config-file/config-advanced` |
| `agents/openai.yaml` keys (`interface`, `dependencies.tools`) | `build-skills` |
| `policy.allow_implicit_invocation` defaults to `true` | `build-skills` |
| Startup loads name, description, and file path | `build-skills` |
| `model_reasoning_effort` ∈ {ultra, max, xhigh, high, medium, low, minimal, none} | `agent-configuration/subagents` |
| `[agents] max_threads` default 6, `max_depth` default 1 | `agent-configuration/subagents` |

All 8 new URLs return 200. All 8 old ones return 308, and each redirect target matches the URL chosen here byte-for-byte — including the `#create-a-skill` anchor and the `?surface=cli` query string.

Unchanged bullets in the prompts, AGENTS.md, and MCP sections were spot-checked and still hold, which is what makes the `Last verified` bump defensible. The MCP field lists remain non-exhaustive (the docs also document `experimental_environment`, `auth`, and `enabled_tools`) — left as-is, since narrowing scope wasn't the goal here.

## Not verified

- **Docs, not behavior.** This is doc-vs-doc throughout; I did not run a Codex binary or read its source. If the docs are wrong, this spec inherits that.
- The `learn.chatgpt.com` pages carry no version stamp, so I can't tell whether a newer Codex release has already moved past them.

## Noted, not fixed

- Line 59 refers to "Copilot's native plugin skills" in a Codex spec — looks like a typo, but it predates this change and I've left it alone.
- `docs/specs/claude-code.md` still contains roughly 25 of the same citation markers. Every other file in `docs/specs/` is already clean, so that's a natural follow-up; happy to send it separately.

## Test plan

Documentation-only — no source, tests, or generated artifacts touched.

- `bun test` — 2610 pass, 0 fail across 97 files
- `bun run release:validate` — "Release metadata is in sync"

### DIFF
--- a/docs/specs/codex.md
+++ b/docs/specs/codex.md
@@ -1,59 +1,64 @@
 # Codex Spec (Config, Prompts, Skills, Subagents, MCP)
 
-Last verified: 2026-04-19
+Last verified: 2026-07-21
 
 ## Primary sources
 
+`developers.openai.com/codex/*` now 308-redirects to `learn.chatgpt.com/docs/*`; the list below reflects the current locations.
+
 ```
-https://developers.openai.com/codex/config-basic
-https://developers.openai.com/codex/config-advanced
-https://developers.openai.com/codex/custom-prompts
-https://developers.openai.com/codex/skills
-https://developers.openai.com/codex/skills/create-skill
-https://developers.openai.com/codex/subagents
-https://developers.openai.com/codex/guides/agents-md
-https://developers.openai.com/codex/mcp
+https://learn.chatgpt.com/docs/config-file/config-basic
+https://learn.chatgpt.com/docs/config-file/config-advanced
+https://learn.chatgpt.com/docs/custom-prompts
+https://learn.chatgpt.com/docs/build-skills
+https://learn.chatgpt.com/docs/build-skills#create-a-skill
+https://learn.chatgpt.com/docs/agent-configuration/subagents
+https://learn.chatgpt.com/docs/agent-configuration/agents-md
+https://learn.chatgpt.com/docs/extend/mcp?surface=cli
 ```
 
 ## Config location and precedence
 
-- Codex reads local settings from `~/.codex/config.toml`, shared by the CLI and IDE extension. citeturn2view0
-- Configuration precedence is: CLI flags → profile values → root-level values in `config.toml` → built-in defaults. citeturn2view0
-- Codex stores local state under `CODEX_HOME` (defaults to `~/.codex`) and includes `config.toml` there. citeturn4view0
+- Codex reads local settings from `~/.codex/config.toml`, shared by the CLI and IDE extension.
+- Configuration precedence, highest to lowest, is: CLI flags and `--config` overrides → project `.codex/config.toml` (nearest the working directory wins) → the profile file selected by `--profile` → user `~/.codex/config.toml` → system `/etc/codex/config.toml` → built-in defaults.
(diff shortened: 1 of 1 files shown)


tmchow:
Closing in favor of deleting docs/specs/codex.md outright. These specs are hand-mirrored snapshots of external docs, referenced by no code or test, and rot silently between re-verifications. Rather than pay the re-verification tax, we're removing the file. Superseded by a deletion PR.

PR #1225, opened 2026-07-22
docs(claude-code): refresh spec for current Claude Code — schemas, skills, hooks

lustrousgorilla:
## Summary

`docs/specs/claude-code.md` was last verified 2026-01-21, making it the oldest spec in the repository by roughly five months. It had drifted across nearly every fast-moving Claude Code surface: the manifest is now optional, commands have merged into skills, plugin agents have a different field contract than documented, hooks have expanded to 30 events and five handler types, and marketplace installation semantics are substantially more precise than the existing text suggested.

This is the source-format spec every converter in `src/converters/` reads from. A wrong claim here does not stay isolated to one target; it can propagate into every conversion path.

## What changed

**Primary sources.** `docs.claude.com/en/docs/claude-code/*` now redirects to `code.claude.com/docs/en/*`. The list uses the current URLs and adds the missing sub-agent and MCP references.

**Plugin layout and manifest.** Documents that `.claude-plugin/plugin.json` is optional, expands the current metadata and component inventories, and distinguishes which custom paths add to defaults, replace defaults, or use component-specific merge rules. It also records validation behavior and version precedence/fallback semantics.

**Commands and skills.** Recasts `commands/` as the legacy flat-skill format, documents the current `SKILL.md` frontmatter inventory, and adds plugin namespacing, frontmatter-name overrides, aliases, invocation visibility, and per-turn tool restrictions.

**Agents.** Replaces the obsolete `capabilities` claim with the current plugin-agent fields and records that plugin agents do not support `hooks`, `mcpServers`, or `permissionMode`.

**Hooks.** Expands the event inventory from 13 to the current 30, adds the `http` and `mcp_tool` handler types, and documents deduplication plus the current plugin/project/data path variables.

**MCP and LSP.** Documents current MCP transports, remote configuration, scoped tool exposure, the full LSP field inventory, and the requirement that language-server binaries be installed separately. The previous MCP `cwd` claim could not be verified in current official docs and was removed.

**Caching and marketplaces.** Narrows cache-copy behavior to marketplace-installed plugins, documents version retention and exact symlink handling, and expands the marketplace schema, source forms, `metadata.pluginRoot`, `defaultEnabled` precedence, and `strict` behavior.

**Citation markers.** Removes 24 leaked browsing-tool artifacts — 75 private-use codepoints (U+E200–U+E202) wrapping tokens like `cite…turn2view0`. Three markers contained multiple source references separated by U+E202, so the cleanup matched complete marker spans rather than assuming a single turn token. No prose was changed in the mechanical stripping pass.

## Verification

Every claim was checked against the live official documentation rather than carried over.

| Claim group | Source |
|---|---|
| Plugin layout, manifest fields, merge rules, validation, versioning, LSP, caching, and symlinks | `plugins-reference` |
| Skill layout, invocation, frontmatter, visibility, and tool restrictions | `skills` |
| Plugin-agent fields, isolation, restrictions, and scoped invocation | `sub-agents` |
| Hook configuration, all 30 events, five handler types, parallelism, and deduplication | `hooks` |
| MCP transports and stdio/remote configuration | `mcp` |
| Marketplace requirements, sources, plugin root, precedence, and strict mode | `plugin-marketplaces` |

All six current primary-source URLs return 200. Legacy `docs.claude.com` locations redirect to their `code.claude.com` equivalents.

An independent reviewer checked the refreshed file against live official documentation without receiving the implementation conclusions. It found four issues — version-fallback scope, marketplace catalog placement, plugin skill alias behavior, and `metadata.pluginRoot` source resolution. All four were corrected, and its second pass was all-clear.

The original contained 75 private-use codepoints across 24 citation markers. The refreshed file contains zero. Stripping the markers from the original into an independent baseline and comparing that baseline with the result leaves only the intended content edits.

## Not verified

- **Docs, not behavior.** This is doc-vs-doc throughout; I did not construct fixtures against a Claude Code binary or inspect its source. If the official documentation is wrong, this spec inherits that.
- The current documentation pages carry no release-wide version stamp, so the verification date records when these claims were checked, not a single Claude Code version they all describe.

## Noted, not fixed

- The refreshed skill and agent field inventories describe Claude Code's source contract. Existing parser and target-converter support for every field was not audited or expanded in this documentation-only change.

## Test plan

Documentation-only — no source, tests, generated artifacts, or release metadata changed.

- `bun test` — 2610 pass, 0 fail across 97 files
- `bun run release:validate` — “Release metadata is in sync”
- `git diff --check` — clean
- PUA scan — 0 codepoints remain

### DIFF
--- a/docs/specs/claude-code.md
+++ b/docs/specs/claude-code.md
@@ -1,67 +1,83 @@
 # Claude Code Plugin Spec
 
-Last verified: 2026-01-21
+Last verified: 2026-07-21
 
 ## Primary sources
 
+`docs.claude.com/en/docs/claude-code/*` now redirects to `code.claude.com/docs/en/*`; the list below uses the current locations.
+
 ```
-https://docs.claude.com/en/docs/claude-code/plugins-reference
-https://docs.claude.com/en/docs/claude-code/hooks
-https://docs.claude.com/en/docs/claude-code/slash-commands
-https://docs.claude.com/en/docs/claude-code/skills
-https://docs.claude.com/en/docs/claude-code/plugin-marketplaces
+https://code.claude.com/docs/en/plugins-reference
+https://code.claude.com/docs/en/skills
+https://code.claude.com/docs/en/sub-agents
+https://code.claude.com/docs/en/hooks
+https://code.claude.com/docs/en/mcp
+https://code.claude.com/docs/en/plugin-marketplaces
 ```
 
 ## Plugin layout and file locations
 
-- A plugin root contains `.claude-plugin/plugin.json` and optional default directories like `commands/`, `agents/`, `skills/`, `hooks/`, plus `.mcp.json` and `.lsp.json` at the plugin root. citeturn2view7
-- The `.claude-plugin/` directory only holds the manifest; component directories (commands/agents/skills/hooks) must be at the plugin root, not inside `.claude-plugin/`. citeturn2view7
-- The reference table lists default locations and notes that `commands/` is the legacy home for skills; new skills should live under `skills/<name>/SKILL.md`. citeturn2view7
+- A plugin is a self-contained directory. Its optional manifest lives at `.claude-plugin/plugin.json`; default component locations at the plugin root include `skills/`, `commands/`, `agents/`, `output-styles/`, `themes/`, `monitors/`, `hooks/`, `bin/`, `settings.json`, `.mcp.json`, and `.lsp.json`.
+- The plugin manifest belongs at `.claude-plugin/plugin.json`; component directories and runtime configuration files belong at the plugin root. A repository that is also a marketplace keeps its catalog at `.claude-plugin/marketplace.json`.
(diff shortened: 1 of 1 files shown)


tmchow:
Closing in favor of deleting docs/specs/claude-code.md outright. This mirrored the Claude Code source-format external docs, was referenced by no code or test, and had drifted ~6 months. Rather than re-verify a mirror nothing reads, we're removing it. Superseded by a deletion PR.

lustrousgorilla:
Hahaha, I feel like an idiot 🤦‍♂️ . I've been working on a more substantial feature fork and a review surfaced the inconsistency... should have sanity checked whether the doc was actually referenced anywhere.

Actually, maybe that's a criterion that'd be worth adding to some new `dead-code` reviewer or folding into the `coherence` or `scope-guardian` reviewers... 🤔 

I'll be more diligent the next time I submit a PR. Btw, thanks for all your hard work on this plugin @tmchow - you've really built something superb.

PR #1238, opened 2026-07-23
feat(ticket-scope): add ce-ticket-scope skill

### LINKED ISSUE #1237: Feature request: ticket-scope — interrogate a work ticket until someone who didn't write it could execute it correctly
## Problem

Compound engineering has strong coverage for documents the team *authors* (ce-brainstorm → ce-plan → ce-doc-review), but most day-to-day work in larger teams arrives as something smaller and messier: a tracker ticket someone else wrote. Those tickets under-specify in two distinct ways, and the agent era makes both *worse*, not better:

**(a) Unwritten intent** — things the codebase cannot contain because they live in heads and chat: which of several valid end states was intended, a constraint like release policy, an assumption nobody validated. "Go read the codebase and figure it out" has been valid for senior engineers for a long time, but is and was very costly in engineer time; now works near-instantly for execution — an agent excavates successfully — but excavation converges *confidently* on **a** present reading, not the intended one, or historical context without arbitrary historical spelunking. Agents inverted the cost curve: execution is cheap, so mis-specification is the expensive part, discovered in review after plausible-but-wrong work exists — including confidently wrong output when multiple stable versions exist and the agent anchors on the wrong version's docs.

**(b) Written-but-undiscovered knowledge** — things that exist in the repo but were never named on the ticket: the maintained skill/runbook that already executes the work (everyone knows where repo skills live - so why is this a problem? not everyone uses the same agents - and not everyone is writing tickets from within the codebase - a Product Owner is defining intent from a chat window, not an IDE), with its preflights and failure patterns. Excavation *can* recover these but demonstrably doesn't when a hand-derived procedure looks complete.

**And the mixed-adoption reality:** when a whole team runs compound engineering, `docs/solutions/` and authored plans largely solve this — my own tickets, produced through the plugin, contain everything I need. But tickets *assigned* to me come from people who haven't adopted it, and they're lacking in non-obvious ways. The tracker ticket is the one interface every team member touches regardless of tooling — which makes it the highest-leverage place to raise the floor, and an on-ramp to the rest of the plugin rather than a competitor to it.

Real incidents behind this request (genericized from a production repo):

1. *(class b)* An agent derived a CI dispatch command from the codebase — every parameter correct — and the run still failed on shared hardware, because the artifact preflight lived in an existing skill the ticket never named. The skill's author asked "are you using the skill?" only after the failure. Why was the skill missed? because it was in Provider A's skill directory, not Provider B (different agents / migrating providers). Trivial to state on the ticket (use and update the skill - which is ce-compound's philosophy)
2. *(class a)* A well-templated ticket (branch field *and* a 7-item DoD checklist) still went sideways: one checklist item contained an unresolved either/or ("backport X, *or* add Y"), the implementer built a sanctioned option, and review required the other branch — four months and two closed PRs later the ticket was canceled with one of its own checkboxes waived.
3. *(class a)* A goal-only ticket ("verify the requirement is met, fix if not") omitted which artifact the check runs against; a repo-based analysis passed while the shipped artifact failed. The work converged within days of the operative facts being written down — in comments, after the rework. The operative baseline — the release branch, not main — was exactly the ambient knowledge the ticket never stated, in a repo with multiple live stable branches.

## Proposal

A skill (working name `ce-ticket-scope`) that interrogates a ticket until it passes one bar:

> Could an engineer (or agent) skilled in the relevant technology, but new to this codebase and its history, produce the **intended** result from this ticket alone — without asking anyone?

**Fields it resolves** (each named for the failure mode leaving it blank produces): baseline/starting context (incl. what acceptance checks run against); observable acceptance criteria (the what, never the how; an either/or inside a criterion is unresolved; must-have vs. severable marked); approach resolution (multiple valid paths → is one required, who decides, redirects recorded as scope changes); out of scope; assumptions each tagged `confirmed` / `validate-before-building` (load-bearing for AI-drafted tickets); unwritten context (a decision recorded as a bare chat link is unresolved — content inline, link as provenance); and existing executable knowledge (does a skill/runbook already execute this? — the class-b field; a guide link is documentation, a skill is operator knowledge).

**Who operates it — two explicit cases, because the questions must reach the right person:**

- **Author-operated** (scoping a ticket before assigning it): interactive interrogation, one question at a time via the platform question tool, following ce-doc-review's conventions. The agent self-answers everything the repo can answer first and marks it "proposed — confirm."
- **Implementer-operated** (the common case per the problem statement — you received the ticket): no blocking loop, because the answers live in the *author's* head and only the author can confirm. The skill self-answers what it can, then emits the residue as a **ready-to-post clarification comment addressed to the author** — a batched, tracker-postable question list. Posting it back is the user's action; v1 does not auto-write to trackers.

**Proportionality rule:** the skill first sizes the blast radius (rework cost if mis-specified) and short-circuits below a threshold — a one-line bugfix ticket gets a one-question pass, not seven fields. Field-resolution rules ("N/A in a field that plausibly needs content is missing, not resolved"; an either/or inside a criterion is unresolved) exist to catch boilerplate fill, the failure mode where templates get satisfied while ambiguity survives (incident 2).

**No-fault framing is a hard rule:** attribute to the artifact, never the author. Blanks are normal; baseline-vs-end-state is genuinely easy to conflate. This is designed to make findings land as tooling rather than criticism (that's been the reception of the private deployment's audits so far — stated as design intent, not a guarantee).

## Proposed sequencing

- **Phase 1 (the initial PR): forward mode** — both operator cases above. Smallest useful surface: read a ticket (GitHub Issues first, matching the plugin's existing surface; tracker specifics isolated so others are additive), interrogate/self-answer, emit the filled ticket or the clarification comment.
- **Phase 2: retro mode** — audit a completed/stalled ticket: each unresolved field becomes a finding with a gap → outcome mapping traced through comments/PR reviews, classified honestly as end-state gap / starting-context gap / **no gap — the ticket was fine, the miss was execution** (the third verdict is what keeps findings credible). Batch audits yield pattern tags and "the single change that would have prevented the most." Retro is a different user (spec-quality analysis vs. live scoping) and heavier tracker access, so it ships second, after forward mode proves adoption.
- **Phase 2 also carries the compounding loop**, because retro is what detects misses: a retro classification appends to a `learned-gaps.md` (the field missed + the one question that would have caught it; saves append good-models). Repo-tracked file; append-only with a dedup rule (same field + same question collapses with a counter); distinct from ce-compound's `docs/solutions/` — a checklist tuner, not solution documentation. Cold start: the skill is fully functional with no file present; the PR would ship a small generic starter set distilled from the incident classes above, not my org's seeded data.

## Scope question for maintainers

1. **Own skill (`ce-ticket-scope`)** — my recommendation. It reuses ce-doc-review's headless-mode and question-tool conventions but isn't a persona review of prose: it's an interrogation/emission workflow that terminates in an improved *ticket in the tracker* (or a postable comment), not a reviewed document.
2. **Extend ce-doc-review** with a `ticket` document classification. Cheaper, but the author-interrogation loop, the implementer-case comment emission, and the retro pattern-tag machinery don't fit the persona-review shape.
3. **Extend ce-plan's intake** — ce-plan already asks clarifying questions when input is underspecified, so this is the cheapest real alternative. The reason I don't recommend it: ce-plan's questions serve *producing a plan for the runner*, and its output is a plan artifact. ticket-scope's output is the corrected ticket/comment *in the tracker*, for whoever picks the work up next (including not-the-runner), plus the retro/audit half, which has no home in intake.

## Offer

I have a working version running privately (tuned to my org's tracker and house template, validated by retroactive audits of real tickets). The field definitions and resolution rules transfer as-is; the tracker tuning and seeded audit data are org-specific and would be stripped per the cold-start note above.

alycda:
Closes #1237 (phase 1).

Adds `ce-ticket-scope`: interrogate a work ticket until someone who didn't write it — human or agent — could produce the **intended** result from the ticket alone.

**What's in phase 1 (forward mode only, per the issue):**
- Seven-field resolution: baseline/starting context, observable acceptance criteria (with the either/or-inside-a-criterion rule and must-have vs. severable), approach resolution with a named decider, out-of-scope, assumptions tagged `confirmed`/`validate-before-building`, unwritten context (decision-by-reference is unresolved), and existing executable knowledge (does a runbook already do this?).
- Operator-case routing: author-operated runs interrogate interactively one question at a time; implementer-operated runs (the common case) never block on questions the session user can't answer — they emit evidence-cited proposed readings plus a paste-ready clarification comment addressed to the ticket's author. The skill never writes to the tracker.
- Proportionality triage: blast radius is sized first; small reversible tickets get one question, not seven fields.
- No-fault framing: findings attribute to the artifact, never the author.

**Explicitly not in this PR (phase 2, per the issue):** retro/audit mode, pattern tags, batch summaries, and the `learned-gaps.md` compounding loop.

**Surfaces touched:** `skills/ce-ticket-scope/SKILL.md`, `docs/skills/ce-ticket-scope.md`, catalog row in `docs/skills/README.md`, root `README.md` inventory row, skill count in `tests/release-metadata.test.ts`.

**Validation:** `bun run test` (2653 pass), `bun run release:validate` (32 skills), `bun run plugin:validate` (strict, both manifests). No bundled scripts; prose follows the admission rules and capability-not-tool tracker references.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ The `compound-engineering` plugin currently ships 31 skills and 0 standalone age
 | [`/ce-work`](docs/skills/ce-work.md) | Execute plans with native or cross-model implementation, durable progress, and transactional host-owned integration |
 | [`/ce-code-review`](docs/skills/ce-code-review.md) | Review code with skill-local reviewer personas |
 | [`/ce-doc-review`](docs/skills/ce-doc-review.md) | Review requirements and plan documents |
+| [`/ce-ticket-scope`](docs/skills/ce-ticket-scope.md) | Interrogate a work ticket until someone who didn't write it could execute it from the ticket alone |
 | [`/ce-debug`](docs/skills/ce-debug.md) | Reproduce failures, trace root cause, fix bugs, and prepare non-trivial fixes for PR |
 | [`/ce-compound`](docs/skills/ce-compound.md) | Document solved problems to compound team knowledge |
 | [`/ce-compound-refresh`](docs/skills/ce-compound-refresh.md) | Refresh stale or drifting learnings |
--- a/docs/skills/README.md
+++ b/docs/skills/README.md
@@ -68,6 +68,7 @@ Invoked when a specific need arises — not part of any chain.
 | [`/ce-debug`](./ce-debug.md) | Find root causes systematically — causal chain gate, predictions, post-fix polish/review, PR handoff |
 | [`/ce-code-review`](./ce-code-review.md) | Structured code review with skill-local reviewer personas, confidence-gated findings, four modes |
 | [`/ce-doc-review`](./ce-doc-review.md) | Review requirements or plan documents using skill-local reviewer personas — coherence, feasibility, product-lens, design-lens, security-lens, scope-guardian, adversarial |
+| [`/ce-ticket-scope`](./ce-ticket-scope.md) | Interrogate a work ticket until someone who didn't write it could execute it from the ticket alone — seven-field resolution, operator-case routing (author-interactive vs. implementer clarification comment), proportionality, no-fault framing |
 | [`/ce-simplify-code`](./ce-simplify-code.md) | Refine recently changed code — reuse, quality, and efficiency review; behavior preservation verified |
 | [`/ce-optimize`](./ce-optimize.md) | Metric-driven iterative optimization loops — three-tier evaluation, parallel experiments, persistence discipline |
 
--- a/docs/skills/ce-ticket-scope.md
+++ b/docs/skills/ce-ticket-scope.md
@@ -1,0 +1,68 @@
+# `ce-ticket-scope`
+
+> Interrogate a work ticket until someone who didn't write it — human or agent — could produce the **intended** result from the ticket alone, then emit the fully specified ticket or a batched clarification comment for its author.
+
+`ce-ticket-scope` covers the artifact most day-to-day work actually arrives as: a tracker ticket someone else wrote. The rest of the core loop specifies work the team *authors* (`ce-brainstorm` → `ce-plan`); this skill raises the floor on work that arrives pre-scoped from outside that loop — which makes it usable, and useful, even when only one person on the team runs the plugin.
+
+It is pure prose with **no bundled script**, so it works verbatim on every supported target.
+
+---
+
+## TL;DR
+
+| Question | Answer |
+|----------|--------|
+| What does it do? | Resolves seven fields — baseline, observable acceptance criteria, approach, out-of-scope, tagged assumptions, unwritten context, existing runbooks — until a new-to-the-codebase engineer or agent could execute from the ticket alone |
+| When to use it | Writing or refining a ticket before assigning it; picking up an under-specified ticket someone else wrote |
+| What it produces | Author-operated: the filled, paste-ready ticket. Implementer-operated: proposed readings with evidence + a paste-ready clarification comment for the author |
+| Skip when | The ticket is small, reversible, and single-file — the skill's own proportionality rule short-circuits to a single baseline question |
+
+---
+
+## Example invocations
+
+```text
+# Pressure-test a ticket you are about to assign
+/ce-ticket-scope is this ready to assign? <paste ticket>
+
+# Scope a ticket you received from someone else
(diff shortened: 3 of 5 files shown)


tmchow:
Thanks for contribution. I don't think this is necessary for the plugin and the models do a great job without a skill for this.

PR #1242, opened 2026-07-23
fix(ce-compound): capture every learning in one run via Phase 0 enumeration

MrGTV-love:
## Problem

`ce-compound`'s **"One learning per run"** rule forces N separate invocations (or N sessions) when a session produced N distinct learnings. The rule conflates two different things:

- **one pipeline pass per learning** — legitimate; it protects fresh tree-grounding, per-doc overlap detection, and prevents drafting-context numbering ("Learning 3") from leaking into written docs; and
- **one invocation per learning** — the handicap; it pushes the enumeration loop onto the user.

## Fix

Move the loop **inside** the skill (Full mode):

- **Phase 0: Learning Enumeration** — enumerate the session's distinct learnings first, highest-value first.
- Run Phases 1–2.5 once per learning against a **fresh per-learning run directory** (so `solution.md` never clobbers between learnings), writing each doc to completion before the next starts — so an intra-run near-duplicate is caught by the normal update-vs-create path.
- Name/title each doc from its own content; resolve cross-references only after all docs exist.

Every protection the original rule guarded is preserved — only the per-invocation boundary is removed. In fact, learning 2's overlap check can now see learning 1's just-written doc, which separate sessions could not.

## Scope / safety

- **+17 / −5**, one file (`skills/ce-compound/SKILL.md`); prose/flow only, no script or schema changes.
- **Single-learning sessions are unchanged** — the loop degenerates to a list of length one (today's single pass).
- **Lightweight mode stays single-pass** by design (the tight-context fallback): it documents the highest-value learning and tips the user to re-run for the rest.

## Motivation

Encountered while using `ce-compound` to capture a multi-learning session: the rule required either batching (which it forbids) or N manual re-invocations. This makes the common "one session, several learnings" case a first-class, single-run flow without weakening any of the grounding guarantees.

### DIFF
--- a/skills/ce-compound/SKILL.md
+++ b/skills/ce-compound/SKILL.md
@@ -25,7 +25,7 @@ Captures problem solutions while context is fresh, creating structured documenta
 /ce-compound mode:headless depth:full [context]        # Full non-interactive run
 ```
 
-**One learning per run.** The workflow's grounding, overlap detection, and cross-referencing all assume a single solved problem. When a session produced multiple distinct learnings, run the skill once per learning, sequentially — each run grounds fresh against the tree. Do not batch several learnings through one run and stitch cross-references between the drafts afterward; drafting-context numbering ("Learning 3") leaking into written docs is the failure this rule prevents.
+**One pipeline pass per learning — the run owns the loop.** The research → assembly → grounding pipeline documents one solved problem at a time, but a single invocation captures every learning the session produced by running that pipeline once per learning, in sequence, inside this run. Full mode enumerates the distinct learnings first (see Phase 0: Learning Enumeration), then executes Phases 1–2.5 once per learning against a fresh per-learning run directory — grounding each against the current tree and writing its doc before the next learning starts, so intra-run near-duplicates hit the normal update-vs-create path. The passes stay sequential rather than fanned-out, so only one learning's subagents are live at a time and context frees between passes. Every doc is named and titled from its own content, never from batch position, and cross-references are resolved only after all docs exist. This preserves what per-learning grounding protected — fresh tree state, per-doc overlap, no "Learning 3" position-labels bleeding into docs — while letting one run capture the whole session. Lightweight mode stays single-pass by design (it exists for tight context): it documents the session's highest-value learning and tips the user to re-run for the rest.
 
 ## CONCEPTS.md bootstrap requests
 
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks for this. I think you've spotted an opp to improve the skill since the old rule really did conflate two different things. 

The failure #1073 was guarding against was parallel batch-drafting and stitching cross-references between half-written drafts, not sequential iteration. Nothing about fresh grounding or per-doc overlap actually needs a process boundary, so moving the loop inside the skill could work well. 

The bonus you called out is good: learning 2's overlap check can now see learning 1's just-written doc, which separate sessions could never do.

Before we merge, there are a few side effects of running this as one long loop instead of N invocations that I'd like worked through. Not nitpicks, they are pretty important.

1. **Orchestrator context accumulation.** "Context frees between passes" is true for the subagents but not for the orchestrator. In Full mode the main turn Reads back each learning's full `solution.md` body, `related.json`, session-history prose, and grounding results, and holds all of it. Loop that N times and it accumulates in one context. Full mode already sits opposite Lightweight because of context pressure, so a 4 or 5 learning run can write the first couple of docs cleanly and then degrade on the rest, which is worse than the old model where each learning started fresh. How can we solve this with a graceful "documented k of n, re-run for the rest" fallback so a big session degrades visibly instead of silently truncating?  Given we can' ttell how big the chunks are we might want to simple solve it by say batching at most 3 docs at a time? 

2. **Session history re-runs per learning.** Phase 1 step 4 is session-scoped, not learning-scoped. As written the loop re-runs the whole discovery plus metadata plus synthesis probe over the same session corpus once per learning. That's wasted wall clock time and up to N synthesis subagents where one would do. It should be hoisted above the loop.

3. **Atomicity and the deferred cross-reference pass.** With N invocations each run is atomic and you re-run the one that failed. An internal loop that dies partway has no defined resume point. Related to that, "resolve cross-references only after all docs exist" is a new step with no owner or mechanism. Today each doc's cross-refs get resolved at write time, before grounding validation gates that doc. A crossref pass appended after the loop is a write-after-write into already-validated tracked docs, which steps outside both the "orchestrator writes one final doc" and "grounding validates before it compounds" invariants. Can you spell out who does that pass, what it writes against, and confirm grounding re-runs on anything it edits? And define what happens if the loop dies before it....

4. **The passive auto-invoke path.** `ce-compound` auto-invokes on "that worked" / "it's fixed", and plenty of people wire it to fire headless after a fix. Previously 1 trigger wrote 1 doc. Now one trigger can fan out into several docs. The non-trivial precondition and your Phase 0 "when unsure, leave it out" help, but a passive trigger spraying a few thin docs is a new failure mode worth a sentence of guardrail.

**On evidence:** the rule you're replacing shipped with skill-creator evals (17/17), and since this inverts that contract we'd want it backed the same way before merge. Totally fine to do it however works for you, including just running it by hand and having your agent hand back receipts. The cases I most want to see are a multi-learning session under real context pressure (does the k-of-n fallback fire cleanly?) and a mid-loop failure (what state does it leave on disk?). Cross-checking on Codex as well as Claude would be ideal since we ship to both (and more).

Last small thing: `docs/skills/ce-compound.md` still describes the single-deliverable model, so it'll need a matching update.  The repo's `AGENTS.md` should have already pointed your agent to do that.

tmchow:
Thanks again for spotting the real conflation here — one pipeline pass per learning vs one invocation per learning. That distinction is right, and the product direction (one run capturing multiple durable learnings) is something we want.

I'm going to close this PR rather than amend it. The remaining design work is large enough that it wouldn't fairly stay as a small follow-up on this branch: orchestrator context caps with a visible k-of-n fallback, hoisting session-history above the loop, dropping the deferred cross-reference pass in favor of resolve-at-write against the live tree, passive/auto-invoke guardrails, skill-doc updates, and behavioral eval coverage for the inverted contract.

We'll take the inspiration into a fresh design pass and open a replacement PR from that. Happy to credit you in the follow-up — feel free to jump in there if you want to co-author.

Closing with appreciation for the contribution.

tmchow:
Closing in favor of a fresh design pass that keeps the multi-learning product direction while addressing the architectural gaps above.

PR #1249, opened 2026-07-24
fix(codex): use a local marketplace source so /plugins lists CE's skills

### LINKED ISSUE #1226: codex cli can not find any skills or hooks or mcp or apps
OpenAI Codex (v0.144.6) 
1、codex
2、/plugins

add market then install but 

› 1. Back to plugins  Return to the plugin list.
  2. Install plugin   Install this plugin now.
     Source           Git · https://github.com/EveryInc/compound-engineering-plugin.git
     Auth             Auth on install
     Skills           No plugin skills.
     Hooks            No plugin hooks.
     Apps             No plugin apps.
     MCP Servers      No plugin MCP servers.

LukeTheoJohnson:
## Summary

Codex now shows Compound Engineering's 31 skills in `/plugins` before installation instead of presenting an empty plugin. The marketplace resolves the plugin from the root of its Git-fetched snapshot, where Codex can read the manifest immediately, rather than treating the same repository as a second Git source that exists only after installation.

Fixes #1226.

## Why the old configuration existed

The Git source was a compatibility workaround, not an arbitrary indirection. When the plugin moved to the repository root in #967, the stable Codex release available at the time rejected `source.path: "./"` and registered the marketplace without enumerating its root plugin. #973 therefore switched the entry to a Git URL and added a guard against the failing configuration.

Codex fixed root-local marketplace sources in openai/codex#28771. That support first reached stable Codex 0.142, released about four and a half hours after #973 merged. On current Codex versions the trade-off has reversed: the root-local source works, while the extra Git source hides path-backed components from the pre-install listing.

Users on Codex 0.141 or earlier need to update Codex before using this marketplace configuration.

## Compatibility

Existing installations are not removed or disabled. They continue loading from the versioned plugin cache; after the marketplace snapshot refreshes, the catalog uses the new co-located source, and installing the plugin again refreshes that cache normally.

The repository's `codex:dev` workflow now recognizes both states during the transition: the new local plugin source inside the official Git marketplace, and the former Git plugin source reported by an existing installation. Local-development documentation now distinguishes a cached marketplace install from the live worktree link supplied by `bun run codex:dev -- local`.

The root plugin still materializes the repository into the plugin cache. Slimming the package layout remains separate work.

## Validation

- Codex CLI 0.145 listed version `3.20.0` from the root-local source and installed all 31 of 31 skills.
- A disposable source transition kept an existing Git-source installation enabled before reinstallation, then refreshed normally through `codex plugin add`.
- `bun test tests/codex-dev.test.ts tests/release-metadata.test.ts` — 57 passed.
- `bun run test` — 2,678 passed.
- `bunx tsc --noEmit`
- `bun run release:validate`
- `bun run plugin:validate`

## Security Disclosure

No security-relevant changes.

## Agent Disclosure

- **Model / reasoning — bulk of the work:** GPT-5 · unknown
- **Model / reasoning — opening this PR:** unknown · unknown

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -7,8 +7,8 @@
     {
       "name": "compound-engineering",
       "source": {
-        "source": "url",
-        "url": "https://github.com/EveryInc/compound-engineering-plugin.git"
+        "source": "local",
+        "path": "./"
       },
       "policy": {
         "installation": "AVAILABLE",
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,11 +24,11 @@ When testing current skill files in Codex, run the repository workflow from the 
 ```bash
 bun run codex:dev -- local    # link this worktree's skills and remove CE plugin installs
 bun run codex:dev -- status   # show local/remote state and checkout provenance
-bun run codex:dev -- remote   # restore the official Git-backed plugin
+bun run codex:dev -- remote   # restore the official marketplace-backed plugin
 bun run codex:dev -- remove   # remove both supported CE installation surfaces
 ```
 
-`refresh` is an idempotent alias for `local`. Local mode manages only the exact `$CODEX_HOME/skills/compound-engineering-local` symlink and Compound Engineering plugin IDs; it must not alter unrelated user skills. The symlink includes modified and untracked files from the selected worktree. Start a new Codex session after switching installation modes. Current Codex versions detect direct skill edits automatically; restart only if an edit does not appear. Do not use this repository itself as a Codex marketplace for local testing: its committed marketplace source points to the public Git repository.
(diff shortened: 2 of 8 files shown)


tmchow:
Thanks for tracking this down. I dug into why the repo had the opposite rule, and your fix is right for current Codex.

The old configuration came from a real compatibility problem. #967 moved the plugin to the repository root. Codex 0.141, which was the stable release when #973 landed, rejected `source.path: "./"`: it registered the marketplace but did not enumerate the root plugin. #973 switched the entry to a Git URL to work around that.

Upstream had merged root-local source support in openai/codex#28771, but it did not reach a stable release until Codex 0.142, about four and a half hours after #973 merged. So the earlier result and the result in this PR were both correct for the Codex versions that produced them. On current Codex, the workaround has become the bug because the extra Git source has no plugin root available for the pre-install component listing.

I pushed follow-up commit 9fc1af2b. It:

- updates `codex:dev` to recognize a local plugin source inside the official Git-fetched marketplace;
- continues accepting the former Git plugin source so existing installations are not classified as drifted during the transition;
- updates the local-development docs, which previously said this checkout's marketplace pointed back to GitHub; and
- adds regression coverage for both source shapes.

I also checked the installed-user transition on Codex 0.145. An existing Git-source installation stayed installed and enabled when the marketplace entry changed to the co-located local source. Nothing is removed or disabled. After the marketplace snapshot refreshes, the corrected catalog is used; running `codex plugin add compound-engineering@compound-engineering-plugin` again refreshes the cache normally.

Validation is clean: 2,678 repository tests, typecheck, release metadata validation, and strict plugin validation all pass. I updated the PR description with the history, compatibility note, and current validation results.

PR #1250, opened 2026-07-24
fix(skills): stop headless runs halting at phase boundaries in Opus 5

kieranklaassen:
## Summary

Autonomous `/lfg` runs stopped partway through. The model would finish a phase, correctly name the remaining steps, and end its turn without doing them. Its closing lines were things like "Returning control to LFG" and "waiting on those before composing" - addressed to a caller that does not exist, because phase transitions load instructions into the same conversation rather than spawning anything.

Eight halts are fixed here. All eight reduce to one error: prose written as if a second party were waiting. A caller to hand back to, helpers to wait for, a next step someone else begins, a consumer who picks up after "done", a recipient for a `Return:`, a user who installs a missing tool, someone who fixes broken credentials - and, once the mandates were cut too far, nobody to say the phases are required at all.

The corpus also drops from 386,034 to 305,115 words (-21%), with 55 unattended defaults added where skills previously asked a question with no fallback.

## Why the diff is mostly deletions

The failing mechanism was documentation of a runtime that had stopped existing. A `Skill` tool call does not spawn a subprocess: it loads instructions inline and the same turn continues. A failing run's own log shows two `Skill` calls and ten subagent calls under a single session id, so helper dispatch is a real boundary and the phase seam is not. Under `claude -p`, an assistant message with no tool call ends the session, which leaves "return a summary to the caller" exactly one expressible form.

| # | what the prose did | what it assumed existed |
|---|---|---|
| 1 | mandated helper dispatch, then waited | helpers the harness had already terminated |
| 2 | returned control across a phase seam | a caller that is the same model |
| 3 | ended a step with no instruction to advance | someone who starts the next one |
| 4 | defined completion and stopped there | a consumer who picks up after "done" |
| 5 | labeled its output envelope `Return:` | a recipient for the return |
| 6 | told the user to install a missing tool, then stopped | a user at the keyboard |
| 7 | told the agent to resolve broken credentials first | someone who can fix credentials |
| 8 | left phase invocation to the model's judgment | (my own over-cut, see below) |

## Design decisions worth reviewing

**Field names are data; prose is not.** Every envelope field, status enum, and greppable terminal marker survives untouched, because three test suites and a live gate depend on them. What went is the justification clause around them. `Documentation complete` stays as a literal string; "so callers can detect completion" does not.

**88 stop sites were deliberately kept.** Sorted by one question: who is actually on the other side. Kept when the user is the subject, when nothing remains for anyone to do, when the roles split across a real runtime boundary, when the site is already an unattended default, or when "stop" bounds an activity rather than the turn. `ce-explain`'s check-in pause and `ce-brainstorm`'s one-question interview are the product, not ceremony; they gained an unattended degradation path instead of a deletion.

**Removing the dispatch floors caused a regression, and the last commit fixes it.** Cutting "you must delegate" was well-supported: vendor guidance says so, and 331 of 332 archived runs never dispatched more than one helper concurrently, which falsifies the wall-clock justification the corpus wrote for its own machinery. But it left nothing saying the pipeline's own phase invocations are mandatory, and a run then shipped a task inline while skipping `ce-plan` and `ce-work` entirely, reporting success with no plan artifact. `4cf86bfe` draws the line: a skill decides its own internal delegation, but whether a step runs at all is not its call.

**Two shipped documents changed, not just skills.** `docs/solutions/skill-design/portable-agent-skill-authoring.md` gains capable-model-first as a conditional deviation, because the guide's own rule says a recurring exception belongs in the guide rather than in 535 local carve-outs - and because this repo has a prior learning concluding the opposite, which deserves to be contradicted explicitly rather than by implication.

## Validation

Measured on a probe brief that traverses every phase boundary, same model and effort throughout.

| | before | after |
|---|---|---|
| pipeline completed | 7 of 12 | **8 of 8** |
| p under the old rate | - | **0.0128 (1 in 78)** |
| output-token spread, same task | 7.12x | 2.00x |
| halt signatures in final message | present | none |

The before column is an A/A control: two identical copies of the pre-change plugin, which completed 7 of 12 with output tokens ranging 21,872 to 155,682. That spread is why the sample size was fixed in advance at 8 runs rather than chosen after seeing results, and why earlier small-sample claims about this pipeline were unreliable.

Separately, a run against a repository with a configured remote exercised the shipping branch that no prior run had taken: the branch pushed, an impossible `gh pr create` was handled with a reasoned skip rather than a halt, and both the plan artifact and the residual-findings record were written. A live pull request with real CI is still unexercised.

`bun run test` 2582 pass / 0 fail, `release:validate` and `plugin:validate --strict` clean. Two test retargets, each net-adding an assertion; no test deleted or weakened.

## Test plan

- `bun run test`, `bun run release:validate`, `bun run plugin:validate`
- New: `tests/no-unreleased-model-names.test.ts` fails if a pre-release model identifier reaches this public repo. It matches the shape of such an identifier rather than a list of names, since a test enumerating them would publish what it protects, and reports file:line only because CI logs are public too.

---


kieranklaassen:
Testing a few hours locally

PR #1273, opened 2026-07-28
feat(routing): add configurable cross-harness execution routes

krafczyk:
## Summary

Compound Engineering users can now select execution harnesses, models, and effort by stable CE role or role class without changing the prompts, personas, permissions, or workflow decisions owned by each skill. One global policy can cover every checkout, projects and explicit task intent can narrow it, and `ce-default` restores the skill's built-in behavior at any layer.

The routing boundary is deterministic and fail-closed: resolution freezes configuration provenance and candidate order before dispatch, adapters must return serving evidence, and strict routes reject unverified or mismatched execution. Preferred routes advance only after a terminal, unintegrated attempt, so fallback cannot duplicate work or silently cross trust boundaries.

| Decision | Reviewer-facing effect |
| --- | --- |
| CE-owned role catalog | Users configure stable implementation, review, reasoning, research, and verification classes while CE retains persona selection. |
| Layered policy | Explicit task intent overrides project policy, which overrides the shared global policy; omitted settings continue to inherit. |
| Resolver/adapter split | The central resolver owns policy and receipts; each skill remains responsible for qualifying and executing its own posture. |
| Native OpenCode authority | Selected waves use process-local opaque handles, canonical attempt locks, and concrete provider/model verification. |
| External CE Work handoff | The durable controller independently resolves public configuration and must match the plugin's comparison identifiers before initialization. |
| Recovery and confinement | Config replacement, routed work, optimization measurement, and read-only peers preserve bounded recovery evidence and reject ambiguous state. |

The large generated portion of this PR keeps each independently installable skill self-contained. Canonical schemas, protocols, dispatch roles, resolver code, and consumer identities are synchronized into skill-local copies and protected by byte-parity and real-conversion tests.

Native OpenCode is the supported, live-assurance path. Converted OpenCode trees and other harness adapters remain experimental compatibility surfaces; unsupported mixed native/external continuation, lost prepared handles, and unavailable identity evidence block rather than claiming routed success.

Session-settled decisions carried from planning: CE retains prompt and persona authority; CE owns role classes with role-level exceptions; one cross-harness global layer is narrowed by project and task policy; resolution is deterministic and non-interactive; `ce-default` can reset inheritance; strict routes require verified serving identity; and the resolver centralizes policy while owning skills execute it (all user-directed).

## Validation

- `bun test --max-concurrency=4`: 3,049 passed, 0 failed across 110 files.
- `bun run release:validate`: passed.
- `bun run routing:check`: passed.
- `git diff --check`: passed.
- Root `CLAUDE.md` symlink check: passed.
- `bun run plugin:validate`: not run locally because the repository-pinned `claude` CLI is unavailable; CI installs that validator.

### Attempt-locked routing snapshots

An attempt-locked snapshot separates deciding *what may run* from reporting *what did run*. The resolver freezes merged policy, provenance, ordered candidates, and a candidate-specific lock before any model receives project material; finalization then accepts evidence only against that lock.

```mermaid
flowchart TB
  A[Global + project + task policy] --> B[Frozen routing snapshot]
  B --> C[Candidate-specific attempt lock]
  C --> D[Adapter dispatch]
  D --> E[Serving evidence]
  E --> F[Canonical finalization]
  F -->|verified| G[Accept]
  F -->|terminal and unintegrated| H[Next declared candidate]
  F -->|ambiguous or mismatched| I[Block]
```

This is preferable here to re-reading mutable configuration between attempts, which could reorder recipients after egress or make recovery authorize a route the original dispatch never selected. CE Work and Optimize persist the corresponding evidence so interrupted runs can recover without trusting caller assertions. Do not use this machinery for a single local call whose target has no fallback, durable state, or identity-sensitive consequence.

### Opaque capability handles

An opaque handle is an unguessable reference to authority held inside the trusted host process. The caller can present the handle but cannot inspect or rewrite the frozen wave, candidate, role, or session it represents.

| Alternative | Consequence |
| --- | --- |
| Caller submits routing fields | Untrusted prose or generated skill code can forge host provenance. |
| Signed portable token | Authority becomes replayable across process lifetime and recovery boundaries. |
| Process-local opaque handle | The host can enforce single-use, exact-instance dispatch and invalidate abandoned state. |

OpenCode's `ce_task_prepare` returns one such handle for a selected wave, and `ce_task` consumes each prepared instance at most once. The deliberate cost is that a plugin-process restart loses the handle and blocks instead of reconstructing authority from weaker external data. Use portable authenticated records instead when restart recovery is a requirement and a durable verifier can safely own the key and replay policy.


krafczyk:
Closing this upstream PR because the configurable routing work will be consumed directly from the maintained fork branch.

PR #1290, opened 2026-07-31
fix(ce-work): pin codex worker reasoning effort to high

### LINKED ISSUE #1289: ce-work codex workers run at default reasoning effort; adapter drops user effort with no override
## Summary

`ce-work`'s cross-model codex route runs implementation workers at the codex CLI's built-in default reasoning effort, with no way for the host or the user to raise it. The claude and grok routes both hardcode `--effort high`; codex is the odd one out.

## Where

`skills/ce-work/scripts/cross-model-work.sh`, `adapter_argv()`:

```sh
codex)
  printf '%s\0' codex exec --ignore-user-config --ignore-rules --ephemeral \
    -s workspace-write -C "$WORKSPACE" --json -o "$RAW_RESULT"
  [ "$(route_model codex)" = auto ] || printf '%s\0' -m "$(route_model codex)"
```

## Why the user can't work around it

- `--ignore-user-config` (correctly) drops `~/.codex/config.toml`, including the user's `model_reasoning_effort` — a user who runs codex at `high`/`xhigh` everywhere gets default-effort workers here.
- The engine config contract forbids CLI flags in `work_engine_preferences`, so there's no sanctioned side-channel.
- The codex CLI has no effort-suffixed model slugs (effort is a separate config key), so the model pin can't carry it.

Meanwhile, the sibling routes make an editorial effort choice already:
- `claude` route: `--effort high` (hardcoded)
- `grok-cli` route: `--effort high` (hardcoded)
- ce-code-review's separate adversarial peer script pins codex at extra-high in its own mapping

## Proposed fix

Pass `-c model_reasoning_effort=high` on the codex route, matching the sibling routes' editorial tier. Verified locally via `--emit-adapter codex` that the argv composes cleanly and `codex exec` accepts the flag. PR to follow.

## Observed

Found while running a real ce-work cross-model run (7 codex worker jobs, model pin `gpt-5.6-sol`): workers completed fine, but effort was silently the CLI default rather than the user's configured tier.

nathanvale:
## Summary

`ce-work`'s codex implementation workers now run at an explicit reasoning-effort tier instead of the CLI's built-in default. The adapter passes `-c model_reasoning_effort=high` on the codex route — the same editorial choice the claude and grok routes already hardcode via `--effort high`.

Fixes EveryInc/compound-engineering-plugin#1289.

## Why the gap existed

`--ignore-user-config` (correctly) drops the user's `~/.codex/config.toml`, taking their `model_reasoning_effort` with it — and the engine-config contract forbids CLI flags in `work_engine_preferences`, so users had no sanctioned way to restore it. Codex was the only route running workers at an unpinned default tier.

## Validation (dogfooded on a real run)

- `--emit-adapter codex` introspection composes cleanly: `codex exec --ignore-user-config --ignore-rules --ephemeral -s workspace-write -C <workspace> --json -o <raw-result> -c model_reasoning_effort=high -`
- Ran a full ce-work cross-model unit through the patched adapter (test-only unit in a real repo, model pin `gpt-5.6-sol`): live worker process argv confirmed `-c model_reasoning_effort=high -m gpt-5.6-sol`; the worker completed, terminalized, and integrated normally with the suite green.
- One behavioral observation worth reviewers' eyes: at high effort the worker produced a ~6-minute silent thinking stretch between output events. It stayed within the documented 600s `CE_PEER_IDLE_SECS` window for `incremental` posture, but repos tuning that window lower may want to know effort lengthens quiet gaps.

---

### DIFF
--- a/skills/ce-work/scripts/cross-model-work.sh
+++ b/skills/ce-work/scripts/cross-model-work.sh
@@ -95,8 +95,11 @@ validate_model_override() {
 adapter_argv() {
   case "$1" in
     codex)
+      # --ignore-user-config drops the user's model_reasoning_effort, so pin the
+      # editorial tier explicitly, matching the claude/grok routes' --effort high.
       printf '%s\0' codex exec --ignore-user-config --ignore-rules --ephemeral \
-        -s workspace-write -C "$WORKSPACE" --json -o "$RAW_RESULT"
+        -s workspace-write -C "$WORKSPACE" --json -o "$RAW_RESULT" \
+        -c model_reasoning_effort=high
       [ "$(route_model codex)" = auto ] || printf '%s\0' -m "$(route_model codex)"
       printf '%s\0' -
       ;;
--- a/tests/skills/ce-work-cross-model-routes.test.ts
+++ b/tests/skills/ce-work-cross-model-routes.test.ts
@@ -206,6 +206,7 @@ describe("ce-work fixed write routes", () => {
     expect(codex).toContain("--ephemeral")
     expect(codex).toContain("-s workspace-write")
     expect(codex).toContain("-C <workspace>")
+    expect(codex).toContain("-c model_reasoning_effort=high")
 
     const claude = emit("claude").stdout
     expect(claude).toContain("--safe-mode")


tmchow:
I added a regression check after reproducing this against base and head. On base, `--emit-adapter codex` omitted any effort override; on this branch it emits `-c model_reasoning_effort=high`. The existing route test now asserts that argv contract, so removing the override will fail CI instead of silently returning Codex workers to the default tier.

Validation on the patched tree: `bun run release:validate`; `bun run test` (2,857 pass, 0 fail); and `git diff --check`. The follow-up commit (`50b522e9`) only adds the regression assertion; the original adapter fix is unchanged.

nathanvale:
Thanks for taking a look at this and getting it in!

PR #1407, opened 2026-08-16
fix(ce-setup): support read-only sandboxes

### LINKED ISSUE #1406: ce-setup health check misreports capabilities in Codex read-only sandbox
## Problem

`skills/ce-setup/scripts/check-health` uses Bash here-strings when parsing capability records:

- the dependency loop around line 302
- the result rendering loop around line 517

Bash implements `<<<` with a temporary file. Under `codex exec --sandbox read-only`, that file creation is denied on macOS. The health script still exits 0, but prints shell errors and reports `Optional capabilities 0/5` even when capabilities are installed.

## Reproduction

Run the shipped `check-health` script through Codex with `--sandbox read-only`. The result includes:

```text
cannot create temp file for here document: Operation not permitted
```

and the capability count is corrupted.

## Proposed fix

Parse each record through process substitution backed by `printf`:

```bash
IFS='|' read -r ... < <(printf '%s\\n' "$record")
```

This preserves field parsing without requiring a writable temporary file. Add a focused guard so `check-health` does not reintroduce here-strings.

I have reproduced the failure with Codex CLI 0.147.0 and validated the two-line replacement in the same read-only sandbox.

ruslan-kurkebayev:
## Summary

Replace the two Bash here-strings in `ce-setup` health reporting with `printf` process substitutions. Bash here-strings require temporary-file creation, so Codex read-only sandboxes could corrupt capability counts while the script still exited successfully.

Add a focused invariant test that prevents the health script from reintroducing temporary-file-backed here-strings.

Closes #1406

## Validation

- `bun test tests/skills/ce-setup-check-health.test.ts` — 54 passed
- `bun run test` — 3,162 passed, 1 skipped
- Codex CLI 0.147.0 with `--sandbox read-only` — health script exited 0, reported 2/5 capabilities, and emitted no temporary-file or permission error
- `bash -n skills/ce-setup/scripts/check-health`
- `git diff --check`

## Security Disclosure

This changes shell parsing but introduces no new command execution, interpolation, or external input surface. Records are still parsed into the same fixed fields; the change only removes the temporary-file requirement.

## Agent Disclosure

- **Model:** Codex CLI · GPT-5

### DIFF
--- a/skills/ce-setup/scripts/check-health
+++ b/skills/ce-setup/scripts/check-health
@@ -299,7 +299,7 @@ capability_missing=0
 results=()
 
 for entry in "${deps[@]}"; do
-  IFS='|' read -r name install_cmd url capability <<< "$entry"
+  IFS='|' read -r name install_cmd url capability < <(printf '%s\n' "$entry")
   capability_total=$((capability_total + 1))
   if command -v "$name" >/dev/null 2>&1; then
     capability_ok=$((capability_ok + 1))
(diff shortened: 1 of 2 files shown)


tmchow:
Nice work 👍

PR #1450, opened 2026-08-18
fix(ce-doc-review): 65% smaller SKILL.md, references carry the rest

tmchow:
`ce-doc-review`'s always-loaded body carried the whole front half of the workflow — mode detection, document intake and classification, every conditional persona's activation signals, and the dispatch payload table — 23KB charged to context on every invocation, and truncated past 8,000 bytes on Codex's Agent Plugins path, which cut the run off inside the persona-selection criteria.

## Size

| | Before | After | Change |
|---|---|---|---|
| `SKILL.md` (CRLF-adjusted) | 22,953 | 7,792 | **66% smaller** |
| Corpus (body + references, md + json) | 304,137 | 306,224 | +2,087 |

Corpus grows by 0.7%: the moves were verbatim, and the growth is the Phase 1 extraction step and the re-gated rules described below.

## What moved where

| Relocated | New home | Loaded |
|---|---|---|
| Mode detection, the non-interactive argument contract and what that mode changes, the full question-tool rules and fallback trigger | `references/modes.md` (new) | mandatory, before anything else |
| Document acquisition per mode, the missing-document gate and its exact failure text, all four classification signal sets, the tie-breaker | `references/document-intake.md` (new) | mandatory, Phase 1 |
| Every conditional persona's activation criteria and the announcement format | `references/persona-selection.md` (new) | mandatory, before selection |
| Payload variable table, document slicing, per-persona model tiering, queueing mechanics, reviewer-failure handling | `references/dispatch.md` (new) | mandatory, before dispatch |

The body keeps what must fire without a read: outcome and done condition, the Setup fence, the never-narrate-a-decision-question rule, the parity-pinned docs-root block and the no-path-only `<root>` resolution, the pre-dispatch readability gate, classify-by-shape with the four type names, native-format edits, the security-lens sensitive-data bound, backpressure-is-not-reviewer-failure, the cross-model egress bounds (`cross_model_review_mode`, unset `CROSS_MODEL_PEERS` means unfiltered not unsanctioned, never silently change a model or recipient), and one pointer per reference at its step.

## Readability pass

The first pass squeezed prose to fit the 8,000-byte bound, which produced sentences a careful reader has to parse twice. Those are rewritten as plain English — one idea per sentence, no fused em-dash chains, no dropped connectives — without compressing anything to make room.

- Setup went from one sentence carrying the conflict rule, the mode scoping, and the no-blocking-question consequence to two sentences that state each in turn; the fence-hygiene rule is now its own paragraph.
- Phase 1's classification rule was a single sentence with a nested parenthetical inside an em-dash chain. It is now three sentences, one per readiness value, with the requirements-only omissions in their own sentence.
- The Interactive-mode and Phase 2 pointers no longer inline a table of contents for the reference they point at; they name what the reference owns and the one behavior that must fire from the body.

One block came out rather than being squeezed: the second cross-model paragraph (peer wave, per-lens slices, poll/reap, which lenses never run cross-model) restated what `references/cross-model-review.md` already owns and is read at that exact step. The body keeps the gate, the pointer, and the three egress bounds. Nothing else moved, and no invariant was dropped.

Body is now 7,792 CRLF-adjusted bytes, 208 under the bound. `bun run test` green (3,197 pass).

## Adversarial cross-model read

Grok-4.6 compared the old body against the new corpus. Its finding was not "text is missing" but **mis-gated** — present, yet only behind a read the path needing it never triggers. Five were real and are fixed in 678d738:

- The ID-bearing HTML sibling-mirror rule sat only in the Phase 1 intake reference, while fixes are applied during synthesis; it now sits beside the native-format rule in `synthesis-and-presentation.md`.
- `{origin_path}` and `{settled_ktds}` were documented as extracted "once during Phase 1", but the Phase 1 reference had no extraction step and personas are explicitly forbidden to recover them — an unfilled slot silently disables provenance-gated technique suppression. `document-intake.md` now performs the extraction and says what breaks without it.
- `synthesis-and-presentation.md`, `walkthrough.md`, and `bulk-preview.md` pointed at SKILL.md sections that moved (the decision primer, the interactive pre-load step); those pointers now name the owning references.

## Tests repointed by load-time

- **Body pins kept:** the `## Interactive mode rules` section and its route into `modes.md`, the never-narrate rule, `bounded parallelism`, backpressure-is-not-reviewer-failure, the walkthrough and bulk-preview pointers, the unified type names and `Product Contract only`, native format, `cross_model_review_mode`, and the "unset means unfiltered" clause.
- **Moved to reference greps:** the per-harness question tools and the numbered-list fallback trigger (`modes.md`), the security-lens activation line (`persona-selection.md`), the payload table, decision primer, model tiering, and queueing mechanics (`dispatch.md`), the classification signals and the argument contract / missing-path text (`document-intake.md`), the ID-bearing mirror rule (`document-intake.md`).

`ce-doc-review` is removed from `OVER_BUDGET` in `tests/codex-skill-prompt-budget.test.ts`. `bun run test` (3,197) and `bun run release:validate` are green.

## Extraction eval

Two scenarios against this branch, read-only, each ending in `FILES_READ:`. A = an implementation-ready unified plan with 11 units, P0/P1 tiers, a non-sensitive internal column rename, and no upstream Product Contract signal. B = a non-interactive HTML review whose path is not on disk, then the same review with the file readable.

| Harness | A: type, team, security-lens bound | A: files read | B: missing-doc gate + non-interactive delivery | B: files read |
|---|---|---|---|---|
| Claude Code (Opus 5) | correct — `unified-plan`, scope-guardian + adversarial justified, security-lens refused on non-sensitive churn | SKILL.md, persona-selection, cross-model-review | correct — zero personas, verbatim failure text, HTML native-format edits | SKILL.md, document-intake, modes |
| Grok CLI (grok-4.6) | correct — same team, same refusal, cross-model gate named | SKILL.md, modes, document-intake, persona-selection, dispatch, cross-model-review | correct — verbatim failure text, Applied vs returned split | SKILL.md, modes, document-intake, synthesis-and-presentation |
| Codex CLI 0.147 | correct — `unified-plan`, adversarial on missing provenance, security-lens refused | SKILL.md, four persona prompts, cross-model-review | correct — stops before Phase 2, requirements-only omissions not flagged | SKILL.md and the intake reference |

Every run followed the body into the references it needed, and the two dangerous paths — dispatching personas at an unreadable document, and activating security-lens (and with it a cross-model peer pass) on ordinary storage churn — were refused in all six cells.

## Does the cap bind the shipping path?

No. Per `docs/specs/agent-plugins.md` the root manifest stays schema-less, so no shipped host truncates today. This is the ratchet goal from `docs/solutions/skill-design/size-driven-skill-restructure.md`, plus 15KB less prompt on every `ce-doc-review` invocation on every host.

## Security Disclosure

No security-relevant changes. The cross-model egress bounds that gate sending document content to a peer model — the `cross_model_review_mode` check, the fixed-route disclosure, and the `CROSS_MODEL_PEERS` filter semantics — stay in the always-loaded body rather than moving into a reference.

## Agent Disclosure

Claude Code · claude-opus-5

---

## Live delegation eval

Real end-to-end runs (not a fake-boundary recognition test): each cell ran `ce-doc-review mode:non-interactive` against `docs/plans/2026-07-09-001-feat-grok-plugin-support-plan.md` in a throwaway clone, with the cross-model peer pass explicitly requested. Subagent dispatch and peer CLI shell-outs were left enabled and graded on subprocess/transcript evidence (a PATH shim logged every child `codex`/`claude`/`grok`/`cursor-agent` invocation), not on the model's own claims.

| Observable | main / Claude Code | PR / Claude Code | main / Codex | PR / Codex |
|---|---|---|---|---|
| Classified `unified-plan` | yes | yes | yes | yes |
| Persona reviewers run | 5 | 5 | 6 | 6 |
| Trio lens activated | adversarial, product-lens | adversarial, product-lens | + security-lens | + security-lens |
| Cross-model peer jobs actually launched | 3 (`codex`) | 3 (`codex`) | 4 (`claude`) | 4 (`claude`) |
| `whole-doc` sweep launched | yes | yes | yes | yes |
| feasibility / coherence / scope-guardian kept off cross-model | yes | yes | yes | yes |
| Non-interactive envelope, no blocking question | yes | yes | yes | yes |
| Writes outside the reviewed doc | none | none | none | none |

**Dispatch-proving excerpts.** Child-process log, PR / Claude Code cell — three real detached `codex exec` peer legs in one wave:

```
codex exec - -C .../xmodel-doc-peer-jx7UZ3 --skip-git-repo-check -s read-only -o .../produ…
codex exec - -C .../xmodel-doc-peer-O5OdVR --skip-git-repo-check -s read-only -o .../adver…
codex exec - -C .../xmodel-doc-peer-4J7Kjq --skip-git-repo-check -s read-only -o .../whole…
```

PR / Codex cell, four `claude` peer legs at the mapped tier: `claude -p --model claude-opus-5 --effort high --permission-mode dontAsk --safe-mode --tools '' --max-turns 15 …` (×4).

Claude-Code transcripts show five `Agent` tool calls per cell on both versions (`coherence-`, `feasibility-`, `product-lens-`, `scope-guardian-`, `adversarial-document-reviewer`). The PR cell additionally reads the five new reference files (`modes.md`, `document-intake.md`, `persona-selection.md`, `dispatch.md`, plus the unchanged `cross-model-review.md`) before dispatch, so the extracted material is being followed, not skipped.

**Findings presentation.** Both versions produced the contract shape — observable-consequence titles, `Consequence if unchanged:` / `Change:` / `Basis:` fields, anchors, a Coverage line naming per-reviewer counts and peer terminal states, and a Restated count. Both Claude cells applied the same 10 safe-auto edits to the same locations in the plan and touched nothing else.

**Verdict: no regression.** The PR preserves persona dispatch, the cross-model wave (per-lens legs + `whole-doc` sweep, correct lens exclusions, one sanctioned route), and the finding-presentation contract on both harnesses. Behavior differences between cells were harness-shaped (Codex additionally activated `security-lens`, on both versions alike), not version-shaped.

**Not exercised.** Codex CLI 0.147 exposes no subagent primitive, so persona reviewers ran inline there on *both* versions — the body's "otherwise run the work inline or serially" path, not the `spawn_agent` path. Interactive mode (routing question, per-finding walk-through, bulk preview) was out of scope for a headless run, as were the `cross_model_review_mode: off` skip path and reviewer-failure/backpressure handling (no reviewer failed and no peer timed out in any cell).


tmchow:
Review round 1 (via `ce-skill-work` respond mode) — c102887:

| item | verdict | owning layer | mechanism |
|---|---|---|---|
| Codex: load interactive references only when interaction is required | Change | loading / placement | the Phases 3-5 pointer lost the interactive-mode condition when it was compressed; restored, and a non-interactive run now explicitly stops at the synthesis envelope |
| Cursor: ID-bearing rule pinned wrongly | Change | test load-time split | the guard pinned the rule only where it is decided (intake); it now asserts it where it fires too (fix application in `synthesis-and-presentation.md`) |

Body 7,999 CRLF-adjusted bytes; `bun run test` (3,197) green.

PR #1452, opened 2026-08-18
refactor(skills): 3,300 bytes off the shared context fence, 220 per dispatch skill body

tmchow:
## Summary

The `## Setup` context fence is byte-identical in fifteen dispatch skills, so every word of its prose is charged to the always-loaded window fifteen times. It stated its conditions at length — the question-override enumerated both mode scopings, the anti-rewrite rule listed four prohibitions as separate clauses, and the rerun policy said twice that there is only one rerun. Same conditions, stated once each.

**1,420 -> 1,200 CRLF-adjusted bytes per copy: 220 bytes off every dispatch skill's body, 3,300 across the corpus.** All fifteen copies stay byte-identical.

This is the corpus-wide lever named in #1449, where the fence was the single largest immovable block left in a skill that could not reach the cap.

## What the prose still says

Every condition survives; none is enumerated twice:

| Condition | Before | After |
|---|---|---|
| Run once, at the start, before any subagent dispatch | stated | unchanged |
| Follow the printed directives | stated | unchanged |
| This skill's user-question rules beat a conflicting directive, and no blocking question is asked | "whether those rules are scoped to a non-interactive mode or apply in every mode, in which case this skill's rules win and no blocking question is asked" | "those rules win whatever their scope, and no blocking question is asked" |
| Run exactly as written, as its own command | stated | unchanged |
| No pipe, filter (`head`/`tail`/`grep`), truncation, or batching | four clauses | one list |
| Header + `CE_CONTEXT_END` detect truncation | stated | unchanged |
| Exactly one verbatim rerun, and only on truncation | "That recovery is the only rerun: otherwise do not rerun it within the same invocation" | "the only rerun" |
| A later invocation runs its own | stated | unchanged |
| No Node runtime -> proceed unchanged | stated | unchanged |

**The bash fence itself is untouched.** It is a known-good command that runs on five hosts, and it is where the interpreter probe lives; shortening a fragile command while moving prose around it would be two changes that can each be defended only separately. The `if`/`else` structure, both diagnostic strings, and the `node`/`nodejs` probe are byte-for-byte what shipped.

## Per-skill effect

| skill | before | after | saved | % of body |
|---|---:|---:|---:|---:|
| `ce-simplify-code` | 7,732 | 7,512 | 220 | 2.8% |
| `ce-retune` | 9,985 | 9,765 | 220 | 2.2% |
| `ce-sweep` | 18,583 | 18,363 | 220 | 1.2% |
| `ce-explain` | 22,368 | 22,148 | 220 | 1.0% |
| `ce-doc-review` | 22,953 | 22,733 | 220 | 1.0% |
| `ce-pov` | 23,682 | 23,462 | 220 | 0.9% |
| `ce-compound-refresh` | 26,544 | 26,324 | 220 | 0.8% |
| `ce-debug` | 32,756 | 32,536 | 220 | 0.7% |
| `ce-optimize` | 43,297 | 43,077 | 220 | 0.5% |
| `ce-ideate` | 51,046 | 50,826 | 220 | 0.4% |
| `ce-brainstorm` | 54,598 | 54,378 | 220 | 0.4% |
| `ce-work` | 55,715 | 55,495 | 220 | 0.4% |
| `ce-code-review` | 57,955 | 57,735 | 220 | 0.4% |
| `ce-compound` | 77,475 | 77,255 | 220 | 0.3% |
| `ce-plan` | 112,795 | 112,575 | 220 | 0.2% |

The percentages are against today's mostly un-restructured bodies and understate the effect. The same 220 bytes is 3.6% of `ce-retune`'s body once #1438 lands (6,041 -> 5,821), and roughly 3-4% of every other restructured body in this sweep.

## Follow-up note: what this brings under the cap on its own

**Nothing, on its own.** Fourteen of the twenty-five `OVER_BUDGET` skills carry the fence; the smallest of them, `ce-retune`, is 9,985 on `main` and lands at 9,765. No skill crosses 8,000 from this change alone.

Its value is the other way round: it is 220 bytes of permanent headroom in every dispatch skill's always-loaded window, taken once instead of fifteen times, and it is the difference between "restructured and just under" and "restructured with room" for the skills already through this sweep. For `ce-debug` — 16,407 after its own restructure — it is nowhere near sufficient, which is why #1449 names it as one of three decisions rather than the decision.

## The docs-root block: measured, not warranted

The other candidate was the `<!-- ce-docs-root -->` block (18 skills, ~920 bytes each, `docs-root-rule-parity` pins it verbatim against `tests/fixtures/docs-root-rule.md`). I measured it rather than assuming: after removing the one genuinely vestigial phrase ("exactly as before"), it is three bullets, each carrying a clause the parity test pins — the `docs` default, the fail-closed rejection, and the sole-namespace rule. The available saving is roughly 19 bytes per copy, about 340 across the corpus, in exchange for editing a fail-closed safety block in eighteen skills at once. Not worth it; left alone deliberately.

## Tests

`skill-context-parity` still holds all four of its substrings — the `SKILL_DIR` anchor line, the `"$SKILL_DIR/scripts/context.mjs"` invocation, `Run the fence exactly as written`, and ``ends with `CE_CONTEXT_END` `` — plus its negative assertion that the interpreter is never hardcoded. No test was changed. `bun run test` (3192 pass, 0 fail) and `bun run release:validate` are green.

`tests/scratch-root-preamble-executes.test.ts` is green and untouched: it executes the *scratch-root* preamble, a different shared block, which this change does not touch.

**Executed, not just string-matched.** The context fence has no executor test, so I extracted the shipped bash block from all fifteen skills, substituted the real skill directory for the `SKILL_DIR` placeholder, and ran each on macOS: 15/15 exit 0, print `=== skill context` as the first line and `CE_CONTEXT_END` as the last. That is exactly the truncation contract the prose describes, verified against the prose that describes it.

## Validation

Two adversarial cross-model reads of the old and new prose against `context.mjs`, one asking whether any behavioral condition was dropped or weakened, one asking whether the new wording is ambiguous for a literal model on a non-Claude host.

| Lens | Model | Result |
|---|---|---|
| Did any behavioral condition get dropped or weakened? | Codex | One real finding, fixed — see below |
| Is the new wording ambiguous for a literal model on a non-Claude host? | Grok | "No real losses or ambiguities: every required constraint survives in the rewrite, including mode-scoped question-rule precedence and the one-rerun truncation recovery." |

Codex found that the first compression had dropped the question-override's **outcome** — "and no blocking question is asked" — and that the replacement, "this skill wins in every mode", reads two ways: precedence regardless of the rule's scope, or a mode-scoped prohibition promoted to global. A literal model on another host could take either, and the second reading is wrong. Restated as the condition plus its outcome — "those rules win whatever their scope, and no blocking question is asked" — which costs 42 bytes per copy of the saving and is worth it. Grok's read was against the corrected text.

## Security Disclosure

No security-relevant changes. The executed shell is byte-identical to what shipped, including the interpreter probe and both failure paths; only surrounding prose changed.

## Agent Disclosure
- **Model:** `Claude Code · claude-opus-5`


tmchow:
Closing: this rewords one shared paragraph in 15 skills for ~220 bytes each, moves nothing to a reference, and the compressed sentences read worse than the originals. Not worth the churn or the merge friction across the open 8KB PRs; the shared fence stays as-is.

PR #1485, opened 2026-08-19
fix(ce-work): require proof quality at the public seam

ethras:
`ce-work` already ran red-before-green for behavior-bearing units. A tautological, internally mocked, or bulk-written test still counted as evidence. This change makes proof quality a gate on that evidence, at execution time.

A test is evidence only when it would fail if the behavior were absent or inverted, and would still pass after a behavior-preserving refactor. The four conditions live in `implementation-loop.md` (the owner) and on the native and cross-model worker packets, because workers write tests before Phase 2 loads the loop.

Seams come from the unit's scenarios and existing test homes. No seam-confirmation question — that would stall `lfg`.

## Validation

- `bun test tests/pipeline-review-contract.test.ts tests/skills/ce-work-outcome-spine.test.ts` — 83 pass
- Skill-creator eval on Claude and Codex: not run in this session (Grok Build host)

## Security Disclosure

No security-relevant changes.

## Agent Disclosure

- **Model:** Grok Build · Grok 4.6


ethras:
Opened against the wrong base (EveryInc/compound-engineering-plugin). This change belongs on the Orca fork (ethras/compound-engineering-orca). Closing; will recreate there.

PR #1489, opened 2026-08-19
fix: honor active model-config keys in skill runs (surface + salience)

### LINKED ISSUE #1488: Skill runs silently ignore active model-config keys (plan_model, cross_model_peer)
## Problem

In a repo with **active, non-commented** model/engine keys in `.compound-engineering/config.yaml`, a full skill run (observed via `lfg`) can silently ignore them and give no signal it did so:

- `plan_model: fable` — `ce-plan` authored the plan on the **session model**, not Fable, though Fable was native and reachable; elevation never fired.
- `cross_model_peer: codex` — `ce-code-review` / `ce-doc-review` **skipped the cross-model pass** ("no sanctioned/authenticated route") **without probing**, though the `codex` CLI was installed and authenticated on `PATH`.

The resolution logic is correct. The failure is **salience asymmetry**:

1. Sibling config (`OUTPUT_FORMAT`) is resolved prominently at `ce-plan` Phase 0.0; model elevation is resolved only at a late, reference-gated Phase 5.2 step that a run can shortcut past.
2. The shared `skills/*/scripts/context.mjs` (15 byte-identical copies) prints `cwd`/`branch`/`head` but nothing about resolved CE config.
3. Pipeline-mode framing ("skip interactive, make choices automatically") reads as "use the session model," even though pipeline only forces `OUTPUT_FORMAT=md` and does **not** disable elevation.
4. The review skills name an "installed route" but never prescribe an actual probe, so unavailability can be asserted rather than observed.

**Impact:** a user who configures a model per skill (a deliberate, sometimes paid choice) silently gets the session model, and the cross-model peer is silently skipped.

## Proposed fix

- **F1** — Surface resolved CE model/engine config in the shared `context.mjs` output (`RESOLVED_CE_CONFIG:`), so the whole class is visible at the top of every run. Minimal scalar reader, no new dependency; applied identically to all 15 byte-identical copies (existing `skill-context-parity.test.ts` guards drift).
- **F2 / F3** — Resolve the model-elevation choice at `ce-plan` / `ce-brainstorm` Phase 0.0, parallel to `OUTPUT_FORMAT`, and state plainly that pipeline mode forces markdown but does not disable elevation.
- **F4** — Require an observed `command -v` (and documented auth) probe of the resolved cross-model peer route before `ce-code-review` / `ce-doc-review` record it unavailable.

Plus edge-case test coverage for the config reader.

Full plan and root-cause analysis are attached in the PR.

kevinold:
## Summary

Configured `plan_model` and `brainstorm_model` choices now become load-bearing at the authoring boundary. The skill surfaces active config at intake, then resolves the current choice immediately before authoring so it cannot silently skip config or override later user intent with stale Phase 0 state.

Fixes #1488.

## Design decisions

- `ACTIVE_CE_CONFIG` reports the raw active value. It does not claim that a shared context reader validated a consumer-specific setting.
- Plan and brainstorm authoring wait until model resolution and any selected dispatch or transparent fallback have settled.
- Provider capability and authentication remain owned by provider-capable dispatch. Early CLI/auth probes were removed because sandbox-local probes can produce false negatives.
- The regression eval isolates the missing config-to-authoring-boundary link. Existing adapter tests separately verify that a selected model reaches dispatch; the eval does not claim which model a live provider served.

## Verification

- `bun run test` — 3,481 tests passed on the clean full-suite run.
- Targeted model-config, reasoning-elevation, review-contract, and eval-catalog tests pass.
- Bounded config-only eval passes on Claude and Codex.
- `bun run release:validate` passes.
- `bun run plugin:validate` passes in strict mode.

## Security Disclosure

Honoring the configured model can activate an existing external-provider route that a skipped gate previously avoided. Existing recipient sanctions, egress controls, credential handling, and fallback rules are unchanged.

## Agent Disclosure

Original implementation by Claude Code · claude-opus-4-8. Follow-up diagnosis, skill-authoring review, and cleanup by Codex · GPT-5.


nish3451 (commented):
review test

nish3451:
**fleet2 PR-LANDING note (2026-08-21):** this PR is blocked on the fork-PR CI approval gate, not on code. Both CI runs (`32311671835`, `32308165775`) are stuck at `action_required` (0s, never executed) pending a maintainer's `Approve and run`. A fleet identity (`nish3451`) is read-only on both this repo and the fork and cannot approve/re-run.

Verified the PR's own CI surface is green locally from the head commit: `bun test` 3419 pass / 1 skip / 0 fail (incl. new `tests/skill-context-config.test.ts` 28/28), `bun run release:validate` in sync, `plugin:validate --strict` passes. No code fix is needed — please `Approve and run` the CI workflow and it should go green.

tmchow:
Pushed `0dd90df9` with the review cleanup:

- label surfaced config as raw `ACTIVE_CE_CONFIG`;
- resolve the latest model choice at the authoring boundary, then block authoring until dispatch or fallback settles;
- remove early auth probes from the non-owning layer;
- add mechanical guards plus a bounded Claude/Codex config-to-gate eval.

The eval proves config reaches model selection; deterministic adapter tests cover dispatch. Full suite (3,481 tests), release validation, and strict plugin validation pass.

PR #1603, opened 2026-08-31
fix(config): probe optional config.local.yaml before Read

### LINKED ISSUE #1593: Skills read optional .compound-engineering/config.local.yaml unconditionally — one phantom Read failure per run in unconfigured repos
Several skills (8+ in v3.21.4: ce-brainstorm, ce-code-review, ce-commit-push-pr, ce-compound, ce-debug, ce-doc-review, and siblings) instruct the model to read options from `<repo>/.compound-engineering/config.local.yaml`. That file is optional by design — most repos never create it — so in an unconfigured repo every one of these skill runs begins with a failed Read.

Nothing breaks: the model shrugs and proceeds with defaults. But the failure is real tooling noise — any harness that captures tool failures (hooks, telemetry, transcript analysis) records one phantom error per skill run per repo, forever. Measured on one busy multi-repo machine: failures across 6+ distinct repos within hours, purely from this probe pattern.

**Suggested fix, one sentence per consumer skill:** "if `.compound-engineering/config.local.yaml` exists (Glob / `test -f`), read options from it; when absent, use defaults without reading." Behavior with a config present is unchanged; unconfigured repos produce zero failed Reads.

Happy to PR the wording change across the consumer skills if that's welcome.

khsaurabh:
Skills that resolve ordinary CE yaml keys used to say **Read** `config.local.yaml` then skip if missing. Models still issued Read, which fails in unconfigured repos (one phantom tool error per skill run).

The shared `ce-config-layers` block now says probe existence (`Glob` / `test -f`) first and do not Read a missing path. Behavior with a present file is unchanged. No new config schema.

Fixes #1593

## Security Disclosure

No security-relevant changes.

## Agent Disclosure

- **Model:** Hermes Agent · grok-4.6

### DIFF
--- a/skills/ce-brainstorm/SKILL.md
+++ b/skills/ce-brainstorm/SKILL.md
@@ -36,7 +36,7 @@ Resolve `<root>` the first time you compose or read a `<root>/` path, never earl
 <!-- ce-config-layers:start -->
 **Resolve ordinary CE yaml keys from the two repo files.**
 
-- **Read** `<repo-root>/.compound-engineering/config.local.yaml`, then `config.yaml` (`<repo-root>` = `git rev-parse --show-toplevel`). Missing files are skipped. Gitignore does not change resolution.
+- **Probe then read** `<repo-root>/.compound-engineering/config.local.yaml`, then `config.yaml` (`<repo-root>` = `git rev-parse --show-toplevel`). Confirm each file exists (`Glob` or `test -f`) before Read. When a file is absent, skip it and continue — do not Read a missing path. Gitignore does not change resolution.
 - **Win** with the first active (non-commented) value. For scalars, empty is unset; an invalid value continues to the next layer, then the skill default. For lists and maps, a present key — including an empty list or map — replaces the whole key.
 - **Do not** use this rule for `docs_root` — that key is `config.yaml` only.
 <!-- ce-config-layers:end -->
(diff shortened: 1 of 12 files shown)


tmchow:
Appreciate the attempt here. But after digging in, the premise doesn't hold you're tryign to fix, so closing it.

The missing-file case is already decided by the block's stated condition: "Missing files are skipped." A failed Read of an optional file IS the existence probe — one tool call that returns "absent" instead of contents. That's a normal agent pattern, and this repo's own conventions treat it as such ("a non-zero exit is a normal state to interpret, not a failure to suppress"). #1593 concedes the dispositive point itself: nothing breaks, the model proceeds with defaults. The harm is telemetry noise, and if hooks or monitoring alert on an expected probe miss, the fix belongs in that filter — not in ten skill files.

The change also costs more than it saves: in an unconfigured repo it's one call either way (a probe miss instead of a Read miss — same count, different status word), and in a configured repo it doubles the calls (probe + Read, per file). Plus it bakes host-specific mechanisms into a cross-host shared block: `Glob` is a Claude Code tool name that's inert vocabulary on Codex/Cursor/Gemini, and `test -f` is POSIX in a plugin that supports PowerShell-shelled contexts — the exact over-prescription pattern the repo's skill-review standard treats as a defect.

Closing this and #1593. If the phantom-error noise is biting in a specific harness, an issue describing that monitoring setup would be worth having — that's where a real fix would live.

tmchow:
Also @saurabhkagent-lab it seems like you have some automated process to crawl the plugin and fix bugs. Generally speaking we don't find that as helpful and actually distracting. Please focus efforts on actual real bugs that are impacting you, it helps much more than drive-by PRs.

Issue #1555, opened 2026-08-27
Allow configurable fallback chains for model elevation


lechatlisse:
## Feature request

Allow model elevation to use an explicitly configured, ordered fallback chain when the preferred model is unavailable.

## Current behavior

In Compound Engineering 3.23.4, CE Plan accepts a single model through `plan_model`. If that model cannot complete the elevated step, the workflow eventually falls back to authoring inline with the session model.

For example, I encountered this sequence:

1. `plan_model: fable` selected Fable.
2. The provider reported that the weekly Fable usage limit had been reached.
3. CE Plan followed its current recovery behavior and continued inline.

That fallback keeps the workflow moving, but it removes the independent model judgment that model elevation was intended to provide.

## Desired behavior

It would be useful to configure an ordered list of acceptable alternatives, including the reasoning effort for each one.

For example:

1. Fable at `high`
2. Opus at `xhigh`
3. Sol at `xhigh`, launched as a fresh-context subagent
4. Stop if none of the configured models is available

The model names should remain aliases such as `fable`, `opus`, and `sol`, allowing the provider or host to resolve the current model generation rather than requiring version-specific IDs.

The exact configuration format is up to the maintainers.

## Expected properties

- Fall back only to models the user explicitly configured.
- Advance to the next model for clearly identified availability or capacity failures, such as an exhausted usage allowance.
- Preserve the same read-only evidence and decision bundle across attempts.
- Allow a fallback candidate to require fresh context so its judgment remains independent from the orchestrating session.
- Report which models were attempted, why each fallback occurred, and which model and effort ultimately served the request.
- Allow the user to choose whether total exhaustion falls back inline or stops the workflow.
- Preserve the existing single-model behavior for users who do not configure a fallback chain.

## Why this would help

In unattended or pipeline-driven runs, silently dropping from an independent planning model to the orchestrating session model can materially change the strength of the planning gate.

An explicit fallback chain would preserve the current transparent, user-controlled model-selection philosophy while making temporary provider limits less disruptive.

PR #1616, opened 2026-09-01
feat: accept model effort level

### LINKED ISSUE #1415: Configurable reasoning effort for ce-plan and ce-brainstorm model elevation
# Configurable reasoning effort for ce-plan and ce-brainstorm model elevation

## Problem

Compound Engineering 3.22.1 lets a checkout select the elevated planning and brainstorming
model with `plan_model` and `brainstorm_model`, but the Claude CLI elevation adapter fixes
reasoning effort at `high`:

```bash
EFFORT="high"   # settled: elevation runs at high effort
```

By contrast, cross-model document and code review now support persistent
`cross_model_effort` configuration. A checkout can therefore select Fable and its effort for
review, but cannot make the equivalent choice for an unusually difficult plan without patching
a versioned plugin cache or bypassing the supported elevation route.

## Requested behavior

Please expose optional planning and brainstorming effort settings, for example:

```yaml
plan_model: fable
plan_effort: high
brainstorm_model: fable
brainstorm_effort: high
```

The exact schema is flexible; a shared `elevation_effort` would also work if separate settings
are not useful.

## Suggested contract

- Preserve `high` as the default so existing behavior is unchanged.
- Support the effort values honored by the resolved model/adapter, including `xhigh` for Fable.
- Use the existing configuration precedence: current-task request, local config, tracked config,
  then the default.
- Reject or visibly skip unsupported explicit values rather than silently substituting.
- Record requested and actual effort alongside the existing requested/served model receipt when
  the adapter can verify it.
- Cover the new settings in `ce-setup`, the config template/reference, and adapter parity tests.

## Use case

We run Compound Engineering from the ChatGPT/Codex desktop app with Codex orchestrating and
implementing, Fable authoring the reasoning-heavy plan step, and Fable providing the independent
adversarial review. `high` is the right standing default. For occasional architecture, security,
tenancy, or migration plans, we would like to request a higher supported effort for the elevated
planning step without changing the rest of the session or maintaining a local plugin patch.

## Existing Claude Code routing limitation

This behavior is already tracked upstream in
[anthropics/claude-code#79562](https://github.com/anthropics/claude-code/issues/79562), an open
bug with a reproduction for `/effort xhigh` switching a Fable session to Opus. Related reports
also document false-positive Fable safety-classifier fallback and session-level stickiness, for
example
[anthropics/claude-code#66657](https://github.com/anthropics/claude-code/issues/66657).

A receipt-verified smoke test on Claude CLI 2.1.233 reproduced the underlying routing behavior:

- `--model fable --effort high` served `claude-fable-5` in the observed run.
- A stream receipt for `--model fable --effort xhigh` explicitly emitted a
  `model_refusal_fallback` event from `claude-fable-5` to `claude-opus-4-8`, categorized as
  `cyber`, even though the prompt only requested a fixed harmless reply and tools were disabled.

This points to Claude Code's Fable classifier-fallback path, not a Compound Engineering defect,
and it means effort alone does not prove which model actually served the request. Compound
Engineering should continue to validate the requested and served model and fail closed on a
family substitution. We are keeping both planning and review at the observed
`fable`/`high` combination for now, while still validating every receipt.

## Environment

- compound-engineering 3.22.1
- ChatGPT/Codex desktop app on macOS
- Claude CLI with Fable available

Related but not duplicate: #1335 added model/effort configuration for cross-model reviews. This
request is limited to `ce-plan` and `ce-brainstorm` model elevation.

### LINKED ISSUE #1565: cross_model_effort rejects codex `max` and `none`, which the Codex CLI accepts
## Summary

On the codex route, `cross_model_effort` accepts five values: `minimal`, `low`, `medium`, `high`, `xhigh`. The Codex API accepts seven. `none` and `max` are missing, so a checkout that pins either one fails the cross-model pass closed on a value the provider would have honored.

## Where

The same case statement, duplicated:

- `skills/ce-code-review/scripts/cross-model-adversarial-review.sh:317`
- `skills/ce-doc-review/scripts/cross-model-doc-review.sh:307`

```sh
codex:minimal|codex:low|codex:medium|codex:high|codex:xhigh) ;;
```

The list is restated in the comment above each case statement, in `skills/guides/configuration.md:51` ("codex `minimal`..`xhigh`"), and in `.compound-engineering/config.example.yaml:124-125` with its byte-identical copy `skills/ce-setup/references/config-template.yaml`. A seventh site pins the belief as a test assertion, which I come back to below.

## Evidence

codex-cli 0.150.1, plugin 3.23.4. Feeding the CLI a junk value makes the API return its full enum:

```
$ codex exec --ignore-user-config --skip-git-repo-check \
    -c model_reasoning_effort=banana -m gpt-5.6-luna --json 'say ok'

{"type":"error","message":"... [ReasoningEffortParam] [reasoning.effort]
 [invalid_enum_value] Invalid value: 'banana'. Supported values are:
 'none', 'minimal', 'low', 'medium', 'high', 'xhigh', and 'max'."}
```

Both missing values complete normally. `max`:

```
{"type":"item.completed","item":{"type":"agent_message","text":"ok"}}
{"type":"turn.completed","usage":{"reasoning_output_tokens":0,...}}
```

and `none`, which returns a completed turn with reasoning off:

```
{"type":"item.completed","item":{"type":"agent_message","text":"ok"}}
{"type":"turn.completed","usage":{"reasoning_output_tokens":0,...}}
```

So the value travels to the provider and gets validated there. The plugin's list is a hand-maintained copy of that enum, and the enum has grown since it was written.

## Why it matters

Failing closed is the right call, and I am not asking you to change it. #1392 decided it ("Fail closed, never substitute. An explicit model or effort the peer cannot honor is a named skip") while closing #1335, and it is the PR that authored this case statement. My problem is upstream of that: the whitelist is what decides the word "cannot," and today it decides wrong. Set `cross_model_effort: max` and you get

```
effort override 'max' not compatible with route 'codex'
```

which reads as a provider limitation.

The exclusion was deliberate rather than accidental, and that is the part worth knowing: `tests/skills/ce-code-review-cross-model-routes.test.ts:1931` encodes the belief in the fail-closed table as

```ts
["codex", "max"],            // not a codex reasoning level
```

That comment was true of some earlier CLI and is not true of 0.150.1. So this is not an oversight next to the claude arm, which already carries `claude:max` at line 316. It is one stale fact, sitting in the validator and asserted again in CI.

Worth separating from any tier policy: this whitelist gates what a checkout may *pin*, not the script's editorial mapping, which stays at `xhigh` either way. `minimal` and `low` already validate, so widening the range changes no default and settles no question about what tier an adversarial peer should run at.

## Proposed fix

Seven surfaces, one belief:

1. `skills/ce-code-review/scripts/cross-model-adversarial-review.sh:317` -- add `codex:none` and `codex:max`
2. `skills/ce-doc-review/scripts/cross-model-doc-review.sh:307` -- the same (a test asserts the two functions stay byte-identical)
3. and 4. the comment above each, which restates the list
5. `skills/guides/configuration.md:51`
6. `.compound-engineering/config.example.yaml:124-125` and its byte-identical copy `skills/ce-setup/references/config-template.yaml`, which AGENTS.md requires be updated together
7. `tests/skills/ce-code-review-cross-model-routes.test.ts:1931` -- the `["codex", "max"]` row has to leave the fail-closed table, or the change lands CI-red

Six hand-maintained copies of one provider enum is what let it go stale in the first place, so the fix that holds is a `bun test` guard pinning the codex list across the validators, their comments, and the docs, the way #1267 paired its knob change with a drift guard. Without it this issue gets refiled the next time OpenAI adds a tier.

One thing worth deciding rather than assuming: `none` turns reasoning off. If you would rather not offer that on a review peer, that is a fine editorial call, but it is a different reason from the one the code gives now, and the comment should say which it is.

A caveat for the config docs, not an argument against the change: at `max` the peer's silent stretches get longer, so anyone who has narrowed the peer budget #1267 turned into one knob should widen it before pinning `max`. (The roughly six-minute quiet gap reported in #1290 was on `ce-work`'s adapter against `CE_PEER_IDLE_SECS`, a different route and a different budget.)

I am happy to open the PR for all seven surfaces plus the drift guard if you want it.

## Out of scope

`ce-work`'s codex adapter (#1289, #1290) and `skills/ce-pov/scripts/cross-model-pov.sh` both hardcode `high` with no config key. Separate surfaces, not part of this ask.

### LINKED ISSUE #1569: ce-work's codex route hard-codes reasoning effort; the review peer's route_effort() takes an override
## Summary

On 3.22.4 the two codex adapters resolve reasoning effort by opposite rules. `ce-code-review` goes through `route_effort()`, which reads `CROSS_MODEL_EFFORT_OVERRIDE` and falls back to a per-route editorial default. `ce-work` writes the value inline in `adapter_argv()` and offers no override path.

Same CLI, opposite policy. I searched open and closed issues and found nobody asking about the difference, so I am asking rather than sending a patch.

I filed #1565 this morning against the review path's effort whitelist. This is the neighbouring question on the work path, one layer down: #1565 is a stale value list inside a knob that exists, and this is the absence of the knob.

## The two paths

`skills/ce-work/scripts/cross-model-work.sh`, `adapter_argv()`:

```sh
codex)
  # --ignore-user-config drops the user's model_reasoning_effort, so pin the
  # editorial tier explicitly, matching the claude/grok routes' --effort high.
  printf '%s\0' codex exec --ignore-user-config --ignore-rules --ephemeral \
    -s workspace-write -C "$WORKSPACE" --json -o "$RAW_RESULT" \
    -c model_reasoning_effort=high
```

`skills/ce-code-review/scripts/cross-model-adversarial-review.sh`:

```sh
route_effort() {   # <route> -> requested effort: the override where the route takes one, else editorial
  if [ -n "${CROSS_MODEL_EFFORT_OVERRIDE:-}" ]; then
    case "$1" in codex|claude|grok-cli) printf '%s' "$CROSS_MODEL_EFFORT_OVERRIDE"; return 0 ;; esac
  fi
  case "$1" in
    codex) printf 'xhigh' ;;
    claude|grok-cli) printf 'high' ;;
```

The review script also pins `M_CODEX="gpt-5.6-luna"` at line 80. So the plugin already runs Luna, and runs it above `high`.

## Why #1289 and #1314 do not cover this

Both reported that the codex work route ran at the CLI default while claude and grok already carried `--effort high`. The ask was parity, not selectability. #1289's author proposed pinning to `high` in his own report and shipped #1290 two hours later. #1314 was pointed at that merge and closed.

The question of whether the value should be choosable has never been put, which is why I am opening this instead of commenting on either.

## Why a user cannot route around it

The three obstacles #1289 listed still stand. `--ignore-user-config` drops `~/.codex/config.toml`, correctly. The engine config contract forbids CLI flags in `work_engine_preferences`. The codex CLI carries effort as a separate key rather than an effort-suffixed slug, so a model pin cannot smuggle one.

## What makes the gap hard to read from outside

The same model writes code at `high` and reviews it at `xhigh`. You closed #1261 by explaining that the luna/xhigh mapping was deliberate, adopted in #1208 after the peer benchmark was re-run with cost priced in API dollars, and written up in `docs/solutions/skill-design/benchmark-review-peer-model-and-reasoning-tier.md`. That reasoning is legible. The work route's `high` has a comment that explains what it matches, not what it was measured against.

Luna is also where the dial moves most. Artificial Analysis puts it at 50.7 on the Intelligence Index at medium effort and 71.4 at max, and max still costs $0.21 per task. Whether `high` is right for implementation I have no data on. My point is narrower: on this model the value is worth choosing, and right now it cannot be.

## The question

Is `high` on the work route a measured editorial decision like the review peer's `xhigh`, or is it the parity default from #1290 that has not been revisited since?

If the first, a pointer to the write-up closes this and I will stop asking.

If the second, I would like to know whether an override hook on the work adapter is welcome, shaped like the one `route_effort()` already implements.

#1567 is open and unreviewed at the time of writing, so I am not leaning on it. If it lands, `max` and `none` become legal on the codex route and a work-side hook inherits a value space that is already settled. If it does not, the question here stands anyway: the review path takes an override today and the work path does not.

## What this is not

Not a request to change the default. `high` can stay the editorial tier for every route. I am asking about the hook, not the number.

## Environment

- compound-engineering-plugin 3.22.4
- codex-cli 0.150.1
- macOS 26.5.2

Observed argv from `bash skills/ce-work/scripts/cross-model-work.sh --emit-adapter codex`:

```
codex exec --ignore-user-config --ignore-rules --ephemeral -s workspace-write -C <workspace> --json -o <raw-result> -c model_reasoning_effort=high -
```

ucirello:
You can now tell CE how hard the model should think when it writes code, when it writes a plan or brainstorm, and when Codex reviews.

If you don't set it, nothing changes (still high). If you pick a value a tool can't use, we skip it. We never quietly pick a different one.

You can also say this in an LFG run, including OpenCode. If you don't say how hard, we use your checkout config, or high.

OpenCode gets this as `--variant`. Codex review now also accepts `max` and `none`.

Fixes https://github.com/EveryInc/compound-engineering-plugin/issues/1569
Fixes https://github.com/EveryInc/compound-engineering-plugin/issues/1415
Fixes https://github.com/EveryInc/compound-engineering-plugin/issues/1565

Not this PR:
- https://github.com/EveryInc/compound-engineering-plugin/issues/1555 (elevation fallback chains)
- https://github.com/EveryInc/compound-engineering-plugin/pull/1567 (open PR that also tries to fix #1565 only)

## Security Disclosure

We check the effort value before handing it to a CLI. Bad values fail with a clear error. No secrets or permission changes.

## Agent Disclosure

- **Model:** OpenCode · xai/grok-4.6#high


.compound-engineering/config.example.yaml
```diff
@@ -72,7 +72,9 @@
 # elevation.
 
 # plan_model: fable              # model alias for the plan-authoring step (no default)
+# plan_effort: high              # optional Claude CLI effort; omit for editorial high
```
ucirello: omit for editorial -> default

.compound-engineering/config.example.yaml
```diff
@@ -72,7 +72,9 @@
...
 # brainstorm_model: fable        # model alias for approach generation (no default)
+# brainstorm_effort: high        # optional Claude CLI effort; omit for editorial high
```
ucirello: omit for editorial -> default

.compound-engineering/config.example.yaml
```diff
@@ -151,6 +153,7 @@
 #     model: composer
 #   - harness: codex
 #     model: "gpt-5.6"
+#     effort: max                 # optional; omit for editorial high
```
ucirello: omit for editorial -> default

docs/guides/ce-brainstorm.md
```diff
@@ -276,7 +276,7 @@ Not from the Phase 4 menu. Software next steps are `ce-plan` or `lfg` (which pla
 
 ## Model elevation
 
-When you want a specific model for the heavy reasoning step, `ce-brainstorm` can generate approaches on that model instead of your session model. Only approach generation is dispatched, with read access so it can verify its brief. The rest of the skill stays on your session model. Name a model in the prompt (`use fable`, `have opus generate these`), or set `brainstorm_model: <model>` in CE config (`config.local.yaml` then `config.yaml`). A prompt request overrides the config key.
+When you want a specific model for the heavy reasoning step, `ce-brainstorm` can generate approaches on that model instead of your session model. Only approach generation is dispatched, with read access so it can verify its brief. The rest of the skill stays on your session model. Name a model in the prompt (`use fable`, `have opus generate these`), or set `brainstorm_model: <model>` in CE config (`config.local.yaml` then `config.yaml`). A prompt request overrides the config key. Optional `brainstorm_effort` pins Claude CLI effort for that elevated step; omit it for editorial `high`.
```
ucirello: omit for editorial -> default

docs/guides/ce-plan.md
```diff
@@ -297,7 +297,7 @@ No. Proof is on the non-software wrap-up menu (save, publish, or both). Software
 
 ## Model elevation
 
-When you want a specific model for the heavy reasoning step, `ce-plan` can author the plan on that model instead of your session model. Only the interpret-findings-then-author step is dispatched, with read access so it can verify its brief. Dialogue and research stay on your session model. Name a model in the prompt (`use fable`, `have opus plan this`), or set `plan_model: <model>` in CE config (`config.local.yaml` then `config.yaml`). A prompt request overrides the config key.
+When you want a specific model for the heavy reasoning step, `ce-plan` can author the plan on that model instead of your session model. Only the interpret-findings-then-author step is dispatched, with read access so it can verify its brief. Dialogue and research stay on your session model. Name a model in the prompt (`use fable`, `have opus plan this`), or set `plan_model: <model>` in CE config (`config.local.yaml` then `config.yaml`). A prompt request overrides the config key. Optional `plan_effort` pins Claude CLI effort for that elevated step; omit it for editorial `high`.
```
ucirello: omit for editorial -> default

docs/guides/ce-work.md
```diff
@@ -250,12 +250,13 @@ work_engine_preferences:
...
   - harness: codex
     model: "gpt-5.6"
+    effort: max
   - harness: claude
 ```
 
 The [central configuration reference](./configuration.md#implementation-routing) explains how this checkout-local default interacts with current-task, session, and project instructions.
 
-Each candidate has a `harness` (`codex`, `claude`, `grok`, `cursor`, or `opencode`) and an optional `model`. Omitting `model` means that harness's configured default. Composer is a model family reached through Cursor, so it is written as `harness: cursor` plus `model: composer`. Keep CLI flags and commands out of config.
+Each candidate has a `harness` (`codex`, `claude`, `grok`, `cursor`, or `opencode`), an optional `model`, and an optional `effort`. Omitting `model` means that harness's configured default. Omitting `effort` keeps editorial `high`. Composer is a model family reached through Cursor, so it is written as `harness: cursor` plus `model: composer`. Keep CLI flags and commands out of config. OpenCode maps `effort` to `--variant`. A token the resolved route cannot honor is omitted with a named reason, never substituted.
```
ucirello: omit for editorial -> default

docs/guides/configuration.md
```diff
@@ -44,11 +44,11 @@ All settings are optional. Commented examples are documentation, not active valu
...
-| [`ce-work`](./ce-work.md), [`lfg`](./lfg.md) | `work_engine_mode`, `work_engine_preferences` | Ordered implementation-author preferences. Mode is `off`, `prefer`, or `require`; each entry has a `harness` and optional `model`. See [Implementation routing](#implementation-routing). |
+| [`ce-plan`](./ce-plan.md), [`ce-brainstorm`](./ce-brainstorm.md) | `plan_model`, `brainstorm_model`, `plan_effort`, `brainstorm_effort` | Model elevation: send the reasoning-heavy step to a named model (e.g. `fable`, `opus`) instead of the session model. Value is a model alias; a prompt request or an orchestrator's `plan_model:<alias>` carrier (e.g. from `lfg`, honored even in pipeline mode) overrides it. Takes effect on every harness — natively where the host serves the model, else via the Claude CLI, else inline. No default (elevation off). Sibling `plan_effort` / `brainstorm_effort` pin Claude CLI effort (`low`..`max`); omit for editorial `high`. A token that route cannot honor is skipped, never substituted. |
```
ucirello: omit for editorial -> default

docs/guides/configuration.md
```diff
@@ -67,10 +67,11 @@ work_engine_preferences:
     model: composer
   - harness: codex
     model: "gpt-5.6"
+    effort: max
   - harness: claude
 ```
 
-Supported harnesses are `codex`, `claude`, `grok`, `cursor`, and `opencode`. Omitting `model` uses that harness's configured default. Composer is a model family reached through Cursor, so request it with `harness: cursor` and `model: composer`.
+Supported harnesses are `codex`, `claude`, `grok`, `cursor`, and `opencode`. Omitting `model` uses that harness's configured default. Composer is a model family reached through Cursor, so request it with `harness: cursor` and `model: composer`. Optional `effort` is a shared-vocabulary token (`none`/`minimal`/`low`/`medium`/`high`/`xhigh`/`max`/`default`); omit it for editorial `high`. A token the resolved route cannot honor is omitted with a named reason, never substituted. Cursor-family work routes honor none. OpenCode maps `effort` to `--variant`.
```
ucirello: omit for editorial -> default

skills/ce-brainstorm/references/reasoning-elevation.md
```diff
@@ -12,7 +12,7 @@ Resolve the per-skill **model choice immediately before adapter selection**, so
...
+3. **Config** — otherwise use the per-skill key: `plan_model` for ce-plan, `brainstorm_model` for ce-brainstorm. Read it the **same way this skill's Phase 0.0 resolves `plan_output` / `brainstorm_output`**: reuse the repo root already resolved, else run `git rev-parse --show-toplevel`, then apply the ordinary-key rule (`config.local.yaml` then `config.yaml`). Reuse the Phase 0.0 reads if still in hand. Ignore commented (`#`-prefixed) lines. A model alias selects it; missing / commented / invalid / no file selects none. Resolve the sibling effort key the same way: `plan_effort` for ce-plan, `brainstorm_effort` for ce-brainstorm. When set to a token the Claude CLI route honors (`low|medium|high|xhigh|max`), export `CE_ELEVATION_EFFORT` on the existing `start` invocation. When unset, the worker keeps editorial `high`. When set to a token that route cannot honor, omit the env, name that skip, and keep editorial `high`. Native in-harness dispatch has no per-call effort flag. There is no `plan_effort:<token>` carrier.
```
ucirello: omit for editorial -> default

skills/ce-plan/references/reasoning-elevation.md
```diff
@@ -12,7 +12,7 @@ Resolve the per-skill **model choice immediately before adapter selection**, so
...
+3. **Config** — otherwise use the per-skill key: `plan_model` for ce-plan, `brainstorm_model` for ce-brainstorm. Read it the **same way this skill's Phase 0.0 resolves `plan_output` / `brainstorm_output`**: reuse the repo root already resolved, else run `git rev-parse --show-toplevel`, then apply the ordinary-key rule (`config.local.yaml` then `config.yaml`). Reuse the Phase 0.0 reads if still in hand. Ignore commented (`#`-prefixed) lines. A model alias selects it; missing / commented / invalid / no file selects none. Resolve the sibling effort key the same way: `plan_effort` for ce-plan, `brainstorm_effort` for ce-brainstorm. When set to a token the Claude CLI route honors (`low|medium|high|xhigh|max`), export `CE_ELEVATION_EFFORT` on the existing `start` invocation. When unset, the worker keeps editorial `high`. When set to a token that route cannot honor, omit the env, name that skip, and keep editorial `high`. Native in-harness dispatch has no per-call effort flag. There is no `plan_effort:<token>` carrier.
```
ucirello: omit for editorial -> default

skills/ce-setup/references/config-template.yaml
```diff
@@ -72,7 +72,9 @@
 # elevation.
 
 # plan_model: fable              # model alias for the plan-authoring step (no default)
+# plan_effort: high              # optional Claude CLI effort; omit for editorial high
```
ucirello: omit for editorial -> default

skills/ce-setup/references/config-template.yaml
```diff
@@ -72,7 +72,9 @@
...
 # brainstorm_model: fable        # model alias for approach generation (no default)
+# brainstorm_effort: high        # optional Claude CLI effort; omit for editorial high
```
ucirello: omit for editorial -> default

skills/ce-setup/references/config-template.yaml
```diff
@@ -151,6 +153,7 @@
 #     model: composer
 #   - harness: codex
 #     model: "gpt-5.6"
+#     effort: max                 # optional; omit for editorial high
```
ucirello: omit for editorial -> default

skills/ce-work/references/cross-model-execution.md
```diff
@@ -92,7 +92,7 @@ When set, `CE_WORK_RUNS_ROOT` is the parent CE Work directory containing all `<r
...
...o the worker. Before prompt construction or external CLI start, the adapter must pass that runner-exported job id and obtain controller `authorize-dispatch` success by calling `unit-workspace.py` `authorize-dispatch`. That success reads the actual runner metadata and exact worker argv, rejecting a shell prefix or substituted adapter before egress, then atomically binds that job id to the exact attempt before egress along with the authorization digest, workspace, packet path and digest, and result directory while revalidating the controller-owned exact route, model, and intermediary contract. A second job for the attempt is refused. A missing or refused handshake, including hand-authored or cross-attempt authorization, refuses egress. Model pins come only from the authorized projection. The adapter's `--emit-adapter` mode remains introspection only and never authorizes production dispatch. Capture the returned job id immediately with `unit-workspace.py` `record-job`, using the controller-returned `attempt_id` verbatim; this idempotently confirms the same validated binding. Never hold one host tool call open for the external runtime and never let the adapter select another recipient.
```
ucirello: reduce this change to be as narrow as possible

skills/ce-work/references/execution-engines.md
```diff
@@ -63,6 +63,7 @@ work_engine_preferences:
 - `work_engine_preferences`: one or more ordered candidate objects
 - `harness`: `codex | claude | grok | cursor | opencode`
 - optional `model`: a model id or family understood by that harness; omission means its configured default
+- optional `effort`: a shared-vocabulary token; omission keeps editorial `high`. A token the resolved route cannot honor is omitted with a named reason, never substituted.
```
ucirello: omit for editorial -> default

skills/ce-work/scripts/cross-model-work.sh
```diff
@@ -95,14 +95,40 @@ validate_model_override() {
   esac
 }
 
+route_effort() {   # <route> -> requested effort: CE_WORK_EFFORT_OVERRIDE where the route honors it, else editorial high
```
ucirello: omit for editorial -> default

.compound-engineering/config.example.yaml
ucirello, line 61: need to add support for plan_harness and brainstorm_harness.

.compound-engineering/config.example.yaml
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

.compound-engineering/config.example.yaml
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

.compound-engineering/config.example.yaml
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

docs/guides/ce-brainstorm.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

docs/guides/ce-plan.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

docs/guides/ce-work.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

docs/guides/configuration.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

docs/guides/configuration.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-brainstorm/references/reasoning-elevation.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-plan/references/reasoning-elevation.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-setup/references/config-template.yaml
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-setup/references/config-template.yaml
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-setup/references/config-template.yaml
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-work/references/execution-engines.md
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-work/scripts/cross-model-work.sh
ucirello: > omit for editorial -> default

Fixed in 9ea0d33 — those comments now say "omit for default" / "keeps the default".

skills/ce-work/references/cross-model-execution.md
ucirello: > reduce this change to be as narrow as possible

Fixed in 9ea0d33 — the start-step addition is now one sentence: export `CE_WORK_EFFORT_OVERRIDE` when the selected preference has a honoured effort.

.compound-engineering/config.example.yaml
ucirello: > need to add support for plan_harness and brainstorm_harness.

This PR is scoped to effort pins on the existing elevation aliases. Adding `plan_harness` / `brainstorm_harness` is a new routing surface (elevation currently falls through native → Claude CLI → session). I do not want to grow this draft into that without a call on whether it belongs here or in a follow-up.

ucirello: > need to add support for plan_harness and brainstorm_harness.

Doing that as a stacked follow-up PR on top of this one, so this draft stays effort/variant only.

skills/ce-work/scripts/unit_workspace_state.py
```diff
@@ -658,14 +662,20 @@ def route_model_allowed(route: str, model: str) -> bool:
 def fixed_route_contract(binding: dict, egress: dict, word: str = "BLOCKED") -> dict:
     if not isinstance(binding, dict) or not isinstance(egress, dict):
         raise Operational(word, "run binding or egress sanction is malformed")
-    expected_binding_fields = {"mode", "target", "model", "source"}
+    expected_binding_fields = {"mode", "target", "model", "source", "effort"}
     if set(binding) != expected_binding_fields:
-        raise Operational(word, "binding must contain exactly mode, target, model, and source")
+        raise Operational(word, "binding must contain exactly mode, target, model, source, and effort")
```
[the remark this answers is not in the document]
ucirello: > normalize a missing `effort` to `null` while continuing to reject other missing or extra fields.

Addressed: 533ca69f treats a missing `effort` as null so old four-field bindings still start. Extra fields are still rejected.

skills/ce-work/scripts/unit_workspace_state.py
```diff
@@ -658,14 +662,22 @@ def route_model_allowed(route: str, model: str) -> bool:
...
     if not isinstance(binding, dict) or not isinstance(egress, dict):
         raise Operational(word, "run binding or egress sanction is malformed")
-    expected_binding_fields = {"mode", "target", "model", "source"}
-    if set(binding) != expected_binding_fields:
-        raise Operational(word, "binding must contain exactly mode, target, model, and source")
+    required_binding_fields = {"mode", "target", "model", "source"}
+    allowed_binding_fields = required_binding_fields | {"effort"}
+    fields = set(binding)
+    if not required_binding_fields <= fields or not fields <= allowed_binding_fields:
+        raise Operational(word, "binding must contain mode, target, model, and source, and may include effort")
```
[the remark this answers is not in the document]
ucirello: > normalize an absent effort to `null` before persistence/comparison.

Addressed: 20d75bb3 now writes `effort: null` when it was omitted, and resume compares the normalized binding so a four-field run can continue with `effort:null`.

skills/ce-work/scripts/unit_workspace_state.py
```diff
@@ -658,14 +668,22 @@ def route_model_allowed(route: str, model: str) -> bool:
...
     if binding.get("mode") not in {"prefer", "require"}:
         raise Operational(word, "binding mode must be 'prefer' or 'require'")
     source = binding.get("source")
     if not isinstance(source, str) or not source or "\0" in source or len(source.encode()) > 256:
         raise Operational(word, "binding source must be a non-empty string of at most 256 bytes")
+    effort = binding.get("effort")
+    if effort is not None and (not isinstance(effort, str) or effort not in BINDING_EFFORT_TOKENS):
+        raise Operational(
+            word,
+            "binding effort must be null or one of: none, minimal, low, medium, high, xhigh, max, default",
```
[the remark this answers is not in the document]
ucirello: > Carry the resolved effort in the controller-owned authorization schema and derive adapter argv from that authenticated value, as is already done for the model.

Addressed: 3f9a4c9e puts `effort_requested` on the authorization artifact. The adapter reads that authenticated value and ignores ambient `CE_WORK_EFFORT_OVERRIDE` once dispatch is bound.

skills/ce-work/references/input-triage.md
```diff
@@ -24,7 +24,7 @@ This skill discovers plans under `<root>/plans/`. Resolve `<root>` when you firs
...
...ing tail. Before the plan path, accept up to two optional carriers in this fixed order: first one compact JSON object prefixed exactly `implementation_engine:`, then one run id prefixed exactly `implementation_run:`. Fully validate and normalize both before any workspace action. The engine object must contain exactly five fields: `mode` is `prefer` or `require`; `target` is `codex`, `claude`, `grok`, `cursor`, `composer`, or `opencode`; `model` is a string pin or `null`; `effort` is a shared-vocabulary token (`none|minimal|low|medium|high|xhigh|max|default`) or `null`; and `source` is a non-empty caller-visible provenance string. The run carrier is accepted only for return-to-caller recovery and must satisfy the safe-id contract above. Reject malformed JSON, missing/extra fields, invalid field types or values, an unsafe run id, an out-of-order carrier, or a duplicate carrier. The entire remaining string is the plan path. A mode token or carrier with no following path is an error; report it instead of treating control data as a bare prompt. Without either optional carrier, the original `mode:return-to-caller <plan-path>` form is unchanged and standing configuration remains eligible.
```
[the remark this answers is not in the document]
ucirello: > normalize an absent `effort` to `null` here while continuing to reject every other missing or extra field.

Addressed: eda82f3f — input triage now requires mode, target, model, and source; omitted `effort` is treated as null.

skills/ce-work/scripts/unit_workspace_state.py
```diff
@@ -712,6 +736,11 @@ def attempt_authorization(
         "harness": contract["harness"],
         "intermediaries": list(contract["intermediaries"]),
         "model_requested": model or contract["default_model"],
+        "effort_requested": (
+            binding.get("effort")
+            if binding.get("effort") in ROUTE_EFFORT_HONOR.get(route, frozenset())
+            else None
+        ),
```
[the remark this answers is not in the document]
ucirello: > Normalize or safely migrate legacy prepared authorizations before comparison so recoverable in-progress units are not stranded.

Addressed: eda82f3f — resume prepare keeps a recorded authorization whose normalized form matches, and the adapter accepts a missing `effort_requested` as null.

skills/ce-plan/references/reasoning-elevation.md
```diff
@@ -12,7 +12,7 @@ Resolve the per-skill **model choice immediately before adapter selection**, so
...
+3. **Config** — otherwise use the per-skill key: `plan_model` for ce-plan, `brainstorm_model` for ce-brainstorm. Read it the **same way this skill's Phase 0.0 resolves `plan_output` / `brainstorm_output`**: reuse the repo root already resolved, else run `git rev-parse --show-toplevel`, then apply the ordinary-key rule (`config.local.yaml` then `config.yaml`). Reuse the Phase 0.0 reads if still in hand. Ignore commented (`#`-prefixed) lines. A model alias selects it; missing / commented / invalid / no file selects none. Resolve the sibling effort key the same way: `plan_effort` for ce-plan, `brainstorm_effort` for ce-brainstorm. When set to a token the Claude CLI route honors (`low|medium|high|xhigh|max`), export `CE_ELEVATION_EFFORT_OVERRIDE` on the existing `start` invocation. When unset, the worker keeps high. When set to a token that route cannot honor, omit the env, name that skip, and keep high. Native in-harness dispatch has no per-call effort flag. There is no `plan_effort:<token>` carrier.
```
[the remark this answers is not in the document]
ucirello: > resolve the sibling effort key whenever elevation is active, independently of whether the model came from live intent, a carrier, or configuration.

Addressed differently: df491e52. Effort is a Claude CLI pin, not a model-source. Deleted it from the config branch; the CLI adapter reads `plan_effort` / `brainstorm_effort` on `start`.

PR #1731, opened 2026-09-16
fix(review): verify PR scope and document provenance

opita04:
PR reviews can currently inspect local commits beyond the requested PR head, and document reviews can treat an origin path or source marker as proof that a premise was accepted. This change requires exact PR identity and independent acceptance evidence before those shortcuts apply.

- Require a clean checkout at the exact PR head, verify fetched head identity, and inspect the actual patch before classifying a PR as trivial.
- Distinguish validated, source-present, unresolved, and greenfield document provenance; preserve scrutiny for new scope beyond an accepted contract.
- Pass identical provenance context to local and cross-model reviewers, and skip peer dispatch when claimed validation lacks verified evidence.

Adapted to the current split-reference workflow. It retains the existing reviewer roster, finding mechanics, and recipient-selection policy.

Validation:
- Focused reruns passed: 91 review-contract tests, 75 cross-model route tests, 72 unified-plan contract tests, and 16 evaluation-catalog tests. The two new provider-stub tests verify rejected unverified provenance and exact context propagation; an additional rerun covers empty evidence references.
- Full `bun test` completed: 4,076 passed / 64 failed. Three stale wording assertions were corrected and pass above; one missing historical evaluation commit was fetched and its catalog passes. The remaining 60 failures were ENOENT errors from the unchanged shared ce-work harness: its afterAll deletes cached seed directories while later files reuse the cached paths. Running all five affected files as separate Bun processes passed all 63 tests. The single-process suite is not claimed green.
- Release metadata validation, shell syntax, and diff whitespace checks passed.
- Fresh-content scenario evaluation covered descendant PR checkouts, major dependency changes, unverified brainstorm markers, and extensions to accepted requirements; follow-up evaluation found no remaining contradictions.
- The generic skill-creator validator rejects the existing `argument-hint` frontmatter field; no entrypoint frontmatter was changed. Repository release validation passes.
- Provider tests used stubs, not live model calls.


tmchow:
@opita04 thanks for the work here. I'm closing this PR. One part of it was right, and it is landing separately.

The fetched-head check closes a real gap. A fork PR can share a branch name with something on `origin`, and the fetch then succeeds against the wrong commit. I pulled that piece into #1753 and credited you there.

The rest reverses decisions that were made on purpose:

- The ancestor check in `scope.md` lets you review your own PR branch with unpushed fixes on top, and the text says so. Requiring an exact head, a clean tree, and no untracked files sends that flow to `pr-remote`, which reviews the stale pushed diff and ignores the fix.
- The `origin:` / `product_contract_source` signal exists to cut review noise. It does not guard against a forged claim, because the document's author is the user. Nothing in the plugin produces a session-handoff reference or a trusted attestation, so `validated` would almost never resolve, and every brainstorm-derived plan would get full premise scrutiny again. The suppression was added to stop exactly that.
- The dependency paragraph lists cases, and this repo's skill standard asks for a condition instead. The judge prompt already says to answer no when in doubt.

For the next PR: non-maintainer PRs need a linked issue first (AGENTS.md, Contribution gate), and the body needs the Security Disclosure and Agent Disclosure sections. Run `bun run test` instead of bare `bun test`. The package script passes `--parallel`, which isolates test files, and that is likely where your 64 failures came from.

If you have a document where origin suppression hid a finding that mattered, open an issue with it attached, and we can design against that case.

Issue #1734, opened 2026-09-17
Generic subagent dispatch cannot select a named worker profile on Devin


GodsBoy:
## Summary

CE skills dispatch generic subagents for reviewers, scouts, researchers, fixers, and similar workers. On Devin (CLI and Devin Local), `run_subagent` selects a worker by named profile, not by raw model id. The two built-in profiles are `subagent_explore` (read-only tools, no `exec`, no file writes) and `subagent_general` (full tool set, always inherits the parent model).

The plugin has no configuration surface that maps a generic CE dispatch to a named profile. Any CE worker that needs more than read-only tools already lands on `subagent_general`, so it runs the parent model at parent cost. A cheaper or stronger profile can only be requested by hand-editing every dispatch site.

## Two authority classes

CE's generic workers split into two authority classes:

- Read-only workers: reviewers, scouts, researchers, validators. They never mutate tracked project content; per-run scratch artifacts (dossiers, reviewer JSON, run logs) stay in this class.
- Write-capable workers: implementers and fixers that mutate tracked project content.

The split matters because a cheaper profile must never silently gain write authority.

## Proposed configuration

Two optional keys under the existing `.compound-engineering/` cascade (`config.local.yaml` then `config.yaml`):

```yaml
# .compound-engineering/config.yaml
subagent_read_profile: ce-explorer    # generic dispatches that do not mutate tracked content
subagent_write_profile: ce-worker     # generic dispatches that do
```

Semantics:

- Profile names are opaque host-defined strings. The plugin never invents, validates, or defaults one.
- The authority class is chosen per dispatch call, from the work the child actually needs.
- Keys apply only where the host's generic dispatch primitive accepts a named worker profile (on Devin, `run_subagent`'s `profile` argument). Typed or registered-agent selectors are out of scope.
- On hosts with no such selector the keys are inert and dispatch is unchanged. Unset keys are a strict no-op everywhere, so existing Claude Code model-selector behaviour is preserved.
- A resolved profile supersedes the surface's model selection for that dispatch class; a profile and a model override are never passed together.
- Unknown, rejected, or unresolvable names follow the surface's ordinary dispatch-failure or degradation path, are named in coverage or degradation notes, and are never reported as having run.

## Distinction from #1301

[...]

PR #1735, opened 2026-09-17
feat: named worker profiles for generic subagent dispatch

GodsBoy:
Refs #1734

## Summary

- Adds two optional config keys, `subagent_read_profile` and `subagent_write_profile`, resolved through the ordinary `.compound-engineering/` two-file cascade (`config.local.yaml` then `config.yaml`). They name the worker profile a generic CE subagent dispatch requests on hosts whose dispatch primitive accepts one: on Devin, `run_subagent`'s `profile` argument; on Devin Local, the same profile locations and semantics apply through the shared agent harness.
- The split is by authority class. Read-class children never mutate tracked project content (per-run scratch artifacts stay read class); write-class children do. The class is chosen per dispatch call, so a cheaper profile can never receive write work.
- The rule ships as a single canonical delimited block embedded in every generic-dispatch-owning skill surface, plus a parity test that sweeps `skills/**` so new dispatch consumers must be classified with a recorded reason.
- Unset keys are a strict no-op on every host: Claude Code and other supported hosts keep their existing model-selector behaviour unchanged, and keys stay inert where no named-profile selector exists.
- A resolved profile supersedes the surface's model selection for that dispatch class and a model override is never passed alongside it. Unknown, rejected, or unresolvable names follow the surface's ordinary dispatch-failure or degradation path and are never reported as having run.
- No profile ships as a default; `swe-2-high` appears only as clearly labelled example material.
- This is distinct from #1301, which concerns model and effort selectors for the high-stakes adversarial code-review pass. This change governs the generic worker dispatch used across most CE skills.

## Verification

- `bun test tests/worker-profile-rule-parity.test.ts`: the canonical block is verbatim and exactly-once in every classified consumer, the clause pins bind each key to its authority class and the local-then-repo cascade order, and the coverage sweep forces every dispatch-instructing file into CONSUMERS or EXCLUDED.
- Focused contract suite: 840 pass / 0 fail across the parity, prompt-budget, setup-health, pipeline, and per-skill contract files.
- `bun run test` (full suite): 4079 pass, 63 fail on the tested tree. 61 failures sit in the `ce-work` unit workspace-controller and cross-model cluster and reproduce identically on a pristine `upstream/main` worktree: the files share a `/tmp/ce-work-repo-template-*` fixture seed that an earlier file's cleanup deletes mid-run, and each file passes in isolation. The remaining 2 were one content-pin regression this diff introduced in `ce-compound-refresh` (a required blockquoted clause dropped during compression), caught by `tests/compound-support-files.test.ts`, fixed and green before push. No suite failure on the pushed head is attributable to this diff.
- `bun run release:validate`: pass (0 agents, 35 skills, 0 MCP servers).
- `bun run plugin:validate`: pass (marketplace and plugin manifests under `--strict`).
- Devin contract: `devin doctor` on Devin CLI 3000.10.31 (Linux x86_64) reports `pass custom subagent profiles 1 profile(s) loaded` for a profile authored to the documented `.devin/agents/<name>/AGENT.md` shape. No live end-to-end Devin dispatch claim is made: runtime profile-to-model attribution is not exposed by the CLI, so verification is the mechanical contract plus `devin doctor`.

## Test plan

- [x] Worker-profile parity, sweep, and clause pins green
- [x] Existing pinned contracts green (review, pipeline, ideate dispatch, reasoning elevation, prefix)
- [x] `release:validate` and `plugin:validate` green
- [x] `devin doctor` green on a documented-shape profile
- [ ] Maintainer CI on this PR

## Non-goals

- No default or recommended profile is shipped; profile names remain operator-defined.
- Typed or registered-agent selectors and the cross-model adversarial review machinery (#1301) are out of scope.
- No runtime model verification is attempted: the Devin CLI does not expose the model serving a profile, so this change claims only the profile request, never which model ran.
- No change to dispatch behaviour when the keys are unset, and no change on hosts without a named-profile selector.

## Security Disclosure

Adds config-gated named-profile selection for generic subagent dispatch. The authority classes are kept strictly separate so a configured profile cannot route write work to a read-class worker; no secrets, permissions, or manifest changes.


tmchow:
Thanks for the PR. I left my thoughts on the shape in #1734: the gap is real, but I'd like a much smaller fix than this. Details and a proposed rework are there.

Issue #923, opened 2026-06-09
extract-skeleton.py silently returns empty for all current Claude sessions (10-line detection cap)


anthonygiuliano:
## Summary

`ce-sessions`' `extract-skeleton.py` silently returns an empty skeleton for **every current Claude Code session**, because its platform auto-detection only inspects the first 10 lines and current Claude session files front-load more than 10 non-message metadata records. This disables session-history synthesis across every skill that depends on it (`ce-sessions`, `ce-compound`'s optional session pass, etc.) — and it fails *silently* (empty output, `parse_errors: 0`), so it reads as "no relevant prior sessions" rather than a bug.

## Root cause

`plugins/compound-engineering/skills/ce-sessions/scripts/extract-skeleton.py`, platform auto-detection loop:

```python
for line in sys.stdin:
    ...
    if not detected and len(buffer) <= 10:   # <-- caps detection to first 10 lines
        obj = json.loads(line)
        if obj.get("type") in ("user", "assistant"):
            detected = "claude"
        elif obj.get("type") in ("session_meta", "turn_context", "response_item", "event_msg"):
            detected = "codex"
        elif obj.get("role") in ("user", "assistant") and "type" not in obj:
            detected = "cursor"
...
handler = handlers.get(detected, handle_codex)   # <-- None falls through to codex
```

Current Claude Code session JSONL front-loads a metadata preamble before the first message. Observed first 10 record types in a real session:

```
1: last-prompt      6: attachment
2: custom-title     7: attachment
3: agent-name       8: attachment
4: mode             9: attachment
5: permission-mode 10: attachment
```

The first `user`/`assistant` record is at **line 13**. Because detection is capped at line 10, it never sees a message record, `detected` stays `None`, and `handlers.get(detected, handle_codex)` falls through to the **codex** handler. Every Claude record then fails the codex checks → `user: 0, assistant: 0, tool: 0, parse_errors: 0` → empty skeleton.

Reproduces on a session that has 141 `user` + 300 `assistant` records — the parser reads all 1212 lines without error and extracts none.

## Fix

Drop the 10-line cap so detection scans until it finds a decisive record (the `if not detected:` guard already short-circuits once set):

```diff
-    if not detected and len(buffer) <= 10:
+    if not detected:
```

Verified locally: the same session goes from `user:0 assistant:0 tool:0` (88-byte empty skeleton) to `user:26 assistant:106 tool:114` (59 KB skeleton with real content).

## Optional hardening

[...]

PR #1736, opened 2026-09-17
fix(ce-compound): read past the resumed-session preamble in extract-metadata

### LINKED ISSUE #1727: ce-compound session-history: extract-metadata.py drops every resumed session at MAX_LINES = 25 and counts the misses as parse_errors (recurrence of #923 in the sibling script)
## Summary

`skills/ce-compound/scripts/session-history/extract-metadata.py` reads only the first
25 lines of each session file, and `try_claude` yields metadata only from a record that
is both `type == "user"` and carries `gitBranch`. In a resumed Claude Code session that
record sits past line 25, so the file is skipped. The skip is then counted as
`parse_errors`, which is a mislabel: every line the extractor read parses cleanly as JSON,
and several of them carry `cwd` and `gitBranch` on other record types.

The visible effect is `ce-compound` reporting "no relevant prior sessions" with a large
`parse_errors` count beside it, which reads as a corrupt-file problem rather than a
coverage one.

This is the same defect class as #923, which fixed the 10-line detection cap in
`ce-sessions/scripts/extract-skeleton.py` by dropping the cap. The sibling script was not
covered by that change.

## Measurement

81 session files in one `~/.claude/projects/<encoded-cwd>/` directory, run through the
shipped script and again with only `MAX_LINES` raised:

```
MAX_LINES = 25  (as shipped)   {"files_processed": 81, "parse_errors": 38}
MAX_LINES = 400                {"files_processed": 81, "parse_errors": 1}
```

37 of the 38 reported parse errors were not parse errors. Locating the first
`type == "user"` record carrying `gitBranch` in each file:

```
record found                         80 of 81
  within the 25-line cap             43
  past the cap, silently dropped     37   (lines 26, 27, 31, 32, 34, 35, 36, 38, 39, 40, 41, 42)
  deepest                            line 42
no such record anywhere               1   (the one genuine miss)
```

## Why the dropped set is the worst possible one

File size only correlates. Long sessions are the ones that get resumed, and a resumed
session opens with a preamble of queue-operation, system and hook records before the first
user turn, which is what pushes that turn past the cap. So the blind spot lands on the
longest and most-resumed sessions, which are exactly the ones most likely to hold the prior
work a history scan is looking for. In two consecutive `ce-compound` runs here the zero was
reported and believed both times.

## Source

`skills/ce-compound/scripts/session-history/extract-metadata.py` at `082c83e`:

- line 19: `MAX_LINES = 25  # Only need first ~25 lines for metadata`
- lines 21 to 43: `try_claude` returns a result only on `obj.get("type") == "user" and "gitBranch" in obj`
- line 382: `if i >= MAX_LINES: break`
- lines 509 and 511: the "no metadata found" branch increments `parse_errors`, and the meta
  record reports it under that name

## Suggested fix

Two independent one-line changes, either of which restores coverage:

```diff
-MAX_LINES = 25  # Only need first ~25 lines for metadata
+MAX_LINES = 200  # Resumed sessions front-load a preamble before the first user turn
```

Scanning until the metadata is found, as #923 did, is the more robust version, since any
cap is a guess about a preamble length nobody controls. The cost is bounded either way
because `try_claude` breaks as soon as it has both fields.

Separately, the counter deserves splitting, because the two conditions have different
remedies and only one of them is a parse problem:

```diff
-        elif error:
-            parse_errors += 1
+        elif error:
+            no_metadata += 1
```

reported as `files_without_metadata` alongside a real `parse_errors`. As it stands the
number sends an investigation to the JSON parser, where there is nothing wrong.

## Environment

Claude Code on macOS, plugin version 3.26.3, marketplace clone at `082c83e`.

costajohnt:
Closes #1727

## Summary

`extract-metadata.py` stopped reading each session file after 25 lines. Resumed Claude Code sessions front-load dozens of non-message records (`queue-operation`, `system`, `hook`, `file-history-snapshot`, ...) before the first `user` record that carries `gitBranch`, so those sessions were never detected and the session-history probe silently dropped them.

Two changes, both in `skills/ce-compound/scripts/session-history/extract-metadata.py`:

1. `MAX_LINES` 25 -> 200, with a comment naming the resumed-session preamble as the reason. Files are already read in full later for `--keyword` scans, so the extra reads are bounded and cheap. I went with the constant bump over scan-until-found because `extract_from_lines` re-iterates `lines` across five platform probes; a lazy scan would mean restructuring all of them. Happy to switch if you prefer that shape.
2. The same miss was reported as a parse error. `process_file` returned the same `(None, filepath)` for "readable but no platform matched" and for `OSError`, so the batch `_meta` line conflated them. `parse_errors` now covers read failures only; the no-match case is reported under a new `_meta.files_without_metadata` count (batch and stdin modes), emitted only when nonzero. `references/session-history.md` only reads `files_processed`, and every existing `parse_errors` assertion still holds.

Same shape as #1082 (issue #923), which dropped the equivalent ten-line cap in `extract-skeleton.py`.

## Validation

- New test `detects Claude sessions whose metadata preamble exceeds 25 records`: 30 preamble records ahead of `fixtures/claude-session.jsonl` in a mktemp dir. Fails on main (`session` undefined, `parse_errors: 1`), passes with the fix (branch `feat/auth-fix` extracted, `parse_errors: 0`).
- New test `counts files with no recognizable metadata separately from parse errors`: single unrecognized JSONL record, expects `parse_errors: 0` and `files_without_metadata: 1`.
- `bun test tests/session-history-scripts.test.ts`: 107 pass, 0 fail (105 before the two new tests).
- `bun run release:validate`: "Release metadata is in sync."
- `bun run plugin:validate`: both manifests "Validation passed".
- `bun run test` (full suite): 4077 pass, 61 fail. All 61 failures reproduce identically on a clean `origin/main` checkout in the same environment (60 in `ce-work-unit-workspace-*` from a shared `ce-work-repo-template-*` mktemp ENOENT when those files run together, 1 in `ce-prototype light-webserver.js` interface binding). None touch session-history.
- Local corpus check: on one machine's `~/.claude/projects/` (1529 files) one session had its first `user`+`gitBranch` record at line index 34 and was dropped and miscounted at cap 25, found at cap 200. Ran on mktemp copies; no transcript content read into anything.

## Security Disclosure

No security-relevant changes.

## Agent Disclosure
- **Model:** `Claude Code · claude-opus-5`. Work done with Claude Code, reviewed and tested by the author.


tmchow:
nice work @costajohnt , thank you!

PR #1737, opened 2026-09-18
feat(wtf): explain the last message or a supplied source plainly

tmchow:
## Summary

When an agent's reply, a plan, or a document does not make sense, you can now type `/wtf` and get it explained in plain language. With no argument it explains the agent's last message. With a file path, a link, a pasted passage, or a pointer such as "the migration part", it explains that instead.

The skill is a short prompt with no workflow, modes, or output template. What it holds the explanation to:

- It explains the source and does not rewrite it. The answer opens with the point, says what it means for the reader, and is shorter than the source. It does not walk a document section by section.
- A plain explanation tends to drop caveats and sound more certain than the source. `wtf` keeps failures, caveats, open questions, and requests. It adds no claims. If re-reading its own message shows that message was wrong, it says so and corrects it.
- A technical term that matters stays and is defined once. Exact file names, commands, and values stay as written.
- When it cannot tell what the user wants explained, it asks one question instead of guessing.

It is not `ce-explain`, which investigates code and history, and not `ce-noslop`, which rewrites prose and keeps every fact.

## Design decisions

**The name has no `ce-` prefix.** `tests/frontmatter.test.ts` says no new entries should be added to the prefix allowlist. This PR adds one anyway, with a written reason in both prefix guards: `/wtf` is what a confused user types, and `lfg` is the precedent. Renaming to `ce-wtf` is a small change if reviewers prefer to keep the rule.

**Manual-only on every host.** The skill sets `disable-model-invocation: true` so the agent never starts re-explaining itself unasked. Codex ignores that flag, so the skill also ships `agents/openai.yaml` with `policy.allow_implicit_invocation: false`. An explicit `$wtf` still works there. No skill or pipeline calls `wtf`, so leaving Codex's catalog costs nothing. The eight existing manual-only skills lack this file; #1738 adds it to them with a guard test.

## Validation

- `bun run release:validate` passes and reports 36 skills. `bun run plugin:validate` passes for both manifests.
- The four test files this change touches pass (1068 tests).
- The full `bun run test` run had 4 failures in files this change does not touch: three timeouts and the `skill-eval-cell` catalog check. All four pass when their files run on their own (92 tests).
- Not run: the fresh-agent eval on Claude and Codex that the new-skill procedure asks for (activation fixtures, a restraint case, a main-path run). How the prose behaves in a real session is untested. This is why the PR is a draft.

## Security Disclosure

No security-relevant changes.

## Agent Disclosure

- **Model:** `Claude Code · claude-fable-5-1`


tmchow:
@codex review

PR #1738, opened 2026-09-18
fix(skills): keep manual-only skills manual-only on Codex

tmchow:
## Summary

The eight manual-only skills could still be started by Codex on its own, because Codex ignores `disable-model-invocation` and reads a per-skill `agents/openai.yaml` instead. Each of them now ships that file with `policy.allow_implicit_invocation: false`, so Codex stops matching them to user intent. Typing `$ce-polish` and the like still works.

The skills are `ce-dogfood`, `ce-polish`, `ce-product-pulse`, `ce-promote`, `ce-retune`, `ce-setup`, `ce-sweep`, and `ce-test-xcode`. The `ce-retune` case is the costly one: its allowlist comment says model-routing it "would let a cheaper request escalate into an expensive measurement program", and on Codex that was possible.

## Why this is safe

With the policy set to `false`, Codex drops the skill from its skill catalog, so no other skill can reach it there (see `docs/solutions/skill-design/invocation-opt-out-flags-block-sibling-skill-invocation.md`). That does not break anything here:

- No skill invokes any of the eight. Every reference from another skill either tells the user to run it or mentions it by name. `lfg` and its references name none of them.
- Claude Code and Grok already refuse sibling calls to these skills because of the frontmatter flag. This change makes Codex behave the same way.
- The Codex converter copies every subdirectory of a skill (`copySkillDir` in `src/utils/files.ts`), so `agents/` ships on that install path as well as through the plugin.

## Guard

A new test in `tests/skill-conventions.test.ts` requires the frontmatter flag and the Codex policy to be set together, in both directions. A skill in `REQUIRED_MODEL_INVOKED_CALLEES` cannot get the file by accident: the file requires the flag, and the existing test forbids the flag on those skills. I removed one file to confirm the guard fails, and it named `ce-polish`.

The new `wtf` skill in #1737 already ships its own `agents/openai.yaml`, so the two PRs pass this guard in either merge order.

## Validation

- `bun run release:validate` passes. `bun run plugin:validate` passes for both manifests.
- `tests/skill-conventions.test.ts` passes (274 tests).
- The full `bun run test` run had 5 failures in files this change does not touch: four timeouts and the `skill-eval-cell` catalog check. All five pass when their files run on their own (65 tests).
- Not run: an install into a real Codex session to watch the skills leave the implicit catalog. The behavior relies on OpenAI's skills documentation and the measurement recorded in the learning linked above.

## Security Disclosure

No security-relevant changes.

## Agent Disclosure

- **Model:** `Claude Code · claude-fable-5-1`


tmchow:
@codex review

Issue #1557, opened 2026-08-27
Expand support for both OpenCode v1 and v2


ucirello:
Right now I can only use compound-engineer-plugin in OpenCode v1. 

It accomplishes that by adding OpenCode v1 slash commands that instructs the model to load a skill. 

OpenCode v2 is in beta, but quickly iterating towards a stable release. 

I would like to be able to use compound-engineering-plugin in OpenCode v2.

ucirello:
Opencode2 seems to be a bit of a moving target right now. Closing this until it's more stable.

PR #1744, opened 2026-09-19
fix(opencode): register skills and commands on OpenCode V2 package installs

### LINKED ISSUE #1752: OpenCode V2: skills and /commands don't appear on package installs
## Problem

Installing Compound Engineering as a package on **OpenCode V2** (`plugins: ["compound-engineering@git+..."]` via `opencode plugin add`, or a manual config entry) loads the plugin but registers nothing: no skills and no `/commands` appear. Local-directory installs (opening the checkout as a project) work, because OpenCode V2 auto-discovers `.opencode/plugins/`.

## Root cause

OpenCode V2 never runs V1 plugin implementations ([migration guide](https://opencode.ai/v2/docs/build/plugins/migrate-v1)). Package installs resolve the package `main` — which pointed straight at `.opencode/plugins/compound-engineering.js` in its **V1-only shape** (a `server()` config hook) — so on V2 the entrypoint loads, does nothing, and every command silently goes missing.

Verified on OpenCode v2.0.8: the installed package cache resolves `main: ".opencode/plugins/compound-engineering.js"`, whose committed shape has no V2 `setup()`.

## Context

This is the concrete V2 gap behind #1557, which was closed while OpenCode V2 was still a beta moving target. V2 is now stable (2.0.8) and in daily use.

## Proposed fix

Ship the documented dual-shape entrypoint: V2 `setup(ctx)` registering every bundled skill through `ctx.skill.transform` and a matching `/command` for each user-invocable skill through `ctx.command.transform`, with the V1 `server()` config hook unchanged — exported from a root `index.js` that both package and local-directory installs resolve. Also document V2's `plugins` config key alongside V1's `plugin`.

PR #1744 implements this with tests covering both shapes.

lucabattistini:
Fixes #1752, closes the OpenCode V2 gap behind #1557).

## Problem

Installing Compound Engineering as a package on **OpenCode V2** (`plugins: ["compound-engineering@git+..."]`) loaded the plugin but registered nothing: no skills and no `/commands` appeared.

OpenCode V2 never runs V1 plugin implementations. Package installs resolve the package `main` — which pointed straight at `.opencode/plugins/compound-engineering.js` in its V1-only shape (a `server()` config hook) — so on V2 the entrypoint loaded, did nothing, and every command silently went missing. Local-directory installs worked by accident: opening the checkout as a project auto-discovers `.opencode/plugins/`, and this machine's V2 session still saw the fix-in-progress working tree.

## Fix

The plugin now exports the documented dual shape from a root `index.js` that both package and local-directory installs resolve:

- **V2** — `setup(ctx)` registers every bundled skill (35) through `ctx.skill.transform` and a matching `/command` for each user-invocable skill through `ctx.command.transform`. Command execution submits the skill-loading prompt into the caller's session, preserving `delivery`.
- **V1** — `server()` returns the same config hook as before (skills path registration + `$ARGUMENTS` command templates), unchanged in behavior.

Frontmatter parsing stays scoped to the leading `---` block, so a `name:` line inside a fenced YAML example in a skill body can't register a bogus command. Both `user-invocable: false` and `slash: false` suppress the `/command`.

Both config keys are now documented: V2's `plugins` and V1's `plugin` (`.opencode/INSTALL.md`, README's OpenCode section).

## Testing

- New `tests/opencode-plugin.test.ts`: dual-shape entrypoint, V2 registration of all 35 skills + commands, command-execute prompt shape, V1 config hook, and no-clobber semantics — 5 tests
- `tests/opencode-plugin-commands.test.ts` updated to exercise the V1 surface (`server()`) of the dual-shape entrypoint — all green
- Full suite run: the only real failures were the 6 pre-existing `opencode-plugin-commands` tests above; the ce-work workspace-controller failures observed in the parallel pass pass in isolation (known bun parallel-worker child-exit issue)
- `bun run release:validate` and `bun run plugin:validate` pass
- Verified live on OpenCode v2.0.8: with the working-tree plugin, skills load and `/ce-brainstorm` executed through the plugin's command registration in this session

## Security Disclosure

No security-relevant changes. The plugin file reads only its own bundled `skills/` directory and registers prompt-redirecting commands.

## Agent Disclosure
- **Model:** OpenCode · GLM-5.3-Flash


tmchow:
@lucabattistini please use the PR template in entirety. You're missing sections. I'd like to include this in the next release.

lucabattistini:
@tmchow i think i've done it, tell me if I need to do anything else :)

Issue #1743, opened 2026-09-19
fix(ce-babysit-pr): normalize corroborated terminal CheckRuns


levineam:
## Problem

`pr-snapshot` preserves `CheckRun.status` literally. GitHub can emit an internally contradictory tuple such as `status: IN_PROGRESS`, `conclusion: SUCCESS`, and non-null `completedAt`, even though the same-head parent workflow run and check suite are both completed/success. The snapshot then holds the PR open forever despite independently corroborated terminal success.

Observed external tuple: check run `105751662961` on jarvOS PR #310; parent workflow `35391814805` and check suite `95853515703` were completed/success on the same head.

## Scope

- Normalize only that contradictory CheckRun state.
- Require non-null terminal conclusion and completedAt **plus** same-head terminal-success agreement from both parent workflow and check suite.
- Preserve fail-closed behavior for missing, mismatched head, non-success conclusion, pending parent, or conflicting parent/suite evidence.
- Add focused positive and negative fixture tests.

## Definition of done

- `pr-snapshot` reports the corroborated check as terminal without weakening ordinary pending detection.
- Tests demonstrate the positive and each mismatched/negative case.
- A risk-targeted review covers false-positive readiness risk.

## Boundaries

Do not modify, rerun, rebase, or merge jarvOS PR #310. Do not change installed plugin cache or activate/install the resulting plugin; downstream installation is a separate follow-up after the source PR lands.

PR #1754, opened 2026-09-21
fix(ce-babysit-pr): corroborate terminal check runs

levineam:
PR snapshots now recognize GitHub's internally inconsistent `IN_PROGRESS`-but-concluded check state only after the exact job, check run, check suite, and workflow run independently confirm the same PR head and result. Other statuses, incomplete evidence, API failures, or any mismatch remain pending.

Related: #1743

Review: a focused Terra risk review found two gaps in the first draft—missing job-record corroboration and accepting non-`IN_PROGRESS` states. This revision verifies the exact Actions job and leaves every other status unchanged.

Validation: the focused regression (`live fetch normalizes`) passes on the rebased head; the full `pr-snapshot` suite and repository test suite were run during the reviewed repair; `bun run release:validate` and `bun run plugin:validate` pass on this head.

## Security Disclosure

No security-relevant changes.

## Agent Disclosure

- **Model:** Codex · GPT-5


tmchow:
nice work @levineam

